PR #3, opened 2023-01-09
Fix compilation on Windows with MinGW

gfgit:
Hello,
This program seems very interesting!
I really like the code style, especially the `InterfaceItem` system which I will study further in the next weeks.

I use MinGW compiler (bundled with Qt package) but in order to build `traintastic-server`, a few fix must be done.
These changes shouldn't affect other systems/compilers because they are all guarded inside `#ifdef`/`#endif`.

Additional changes to the installer would be needed to ship MinGW runtime for MinGW builds.
If you want I can write a separate Pull Request for that.

**NOTE**:
I used **Qt  5.15.2 64-bit** version but the shipped `MinGW (v. 8.1.0)` has a bug in `std::filesystem` header.
To fix the issue I used the newer version `MinGW 11.2.0` (shipped with Qt 6).
When using tools like `windeployqt` pay attention to put the updated version of `libstdc++6.dll` inside program folder.

Please let me know your thoughts on this matter.
Filippo

### DIFF
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -210,6 +210,13 @@ endif()
 
 ### LIBRARIES ###
 
+if(WIN32 AND NOT MSVC)
+  # On MinGW ws2_32.dll needs explicit linking
+  # Also mswsock.dll is needed for AcceptEx() used by Boost.Asio
+  target_link_libraries(traintastic-server PRIVATE ws2_32 mswsock)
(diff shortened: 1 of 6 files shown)


gfgit:
Currently I have not tested it with Qt 6 but I will install it because I'd like to port my program to Qt6 too.

reinder:
Looking good, I'll merge it :)

PR #5, opened 2023-01-13
input: fix TraintasticDIY input simulation

gfgit:
Toggling a disabled input caused crash.
Prevent iterator derefencing when invalid.

Other interfaces, like z21, allow toggling disabled inputs so the behaviour is still correct.


gfgit:
This also fixes toggling of enabled inputs.
When inputs where in state `True`, they wouldn't go back to `False` because of the `&&`

reinder:
Thanks for the fix :)

PR #13, opened 2023-01-19
Server: handle SIGINT, SIGTERM, SIGBREAK

### LINKED ISSUE #10: Gracefully shutdown from command line
**Describe the bug**
When not running with `--tray` there is now elegant way to close the server CMD window.
`CTRL-C` and `CTRL-Z` do not work.
Maybe add [control-handler-function](https://learn.microsoft.com/en-us/windows/console/registering-a-control-handler-function) (Windows specific).
Or catch `SIGINT or SIGTERM`signal (Unix specific, may work on Windows too)
Or maybe add both.
Crashing at shutdown prevents auto-saving.

gfgit:
Register signal handlers to shutdown gracefully

Closes #10


reinder:
Looking good already!, some comments:
- There is already signal handling for Unix implemented, that must be removed.
- The log message is missing, see [server/src/os/unix/signals.cpp](https://github.com/traintastic/traintastic/blob/master/server/src/os/unix/signals.cpp)
- Did you test it on Linux and/or macOS?

p.s. please add a space between `//` and the comment text :)

reinder:
You can disable clang-tidy, that will make it a bit faster. On the GitHub action runners a full rebuild takes about 20min on two cores.

On which distro are you compiling?

Just Linux is fine, macOS and Linux are both Unix, so I'd expect it to work.

gfgit:
Kubuntu 22.10

reinder (approved):

gfgit:
The PR wasn't ready yet :(
I forgot to remove `os/unix/signals.hpp` header from `main.cpp`

gfgit:
@reinder

PR #17, opened 2023-01-22
Windows Tray: set DPI aware

gfgit:
Prevent blur on Windows High DPI Monitor.
Tested on Windows 10, 1920x1080.

This might drop support for older version on Windows (like XP)
[Microsoft Docs](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setprocessdpiaware) suggest to add a manifest file to the executable instead of manually calling this function.
Qt 5.15 already doesn't support Windows XP so client cannot run on it but server could.
Maybe by adding a manifest file instead of calling this function, older Windows version might just ignore the value and run normally.

### DIFF
--- a/server/src/os/windows/trayicon.cpp
+++ b/server/src/os/windows/trayicon.cpp
@@ -29,6 +29,8 @@
 #include "../../traintastic/traintastic.hpp"
 #include "../../utils/setthreadname.hpp"
 
+#include <iostream>
+
 namespace Windows {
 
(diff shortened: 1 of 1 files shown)


reinder:
Traintastic will only support W10 and above, so no need to support old windows versions.

reinder:
I think we should go with the manifest resource, as that is recommended by Microsoft.

gfgit:
I agree

PR #18, opened 2023-01-22
Fix gcc warnings

gfgit:
Fix some GCC warnings.
I noticed there are a lot of unused `#include ...` maybe it's worth running a dedicated tool to find all of them


gfgit:
There are also some warnings about classId in InputMonitor and OutputKeyboard constructors but I don't know how to change them

reinder:
includes of project files are probably left overs of some refactoring, it would indeed be good to do some clean up.
standard library includes are a bit tricky, removing them for one compiler may work, but not for others.

PR #23, opened 2023-01-26
Add Italian translation

### LINKED ISSUE #20: Add Italian translation

gfgit:
Closes #20

- Before merging I'd like to remove old translations and add missing ones.
- Fixed a typo in English translation (can be applied from different PR if you prefer)
- Other language need to be update to show Italian language and also installer needs to be update too.
- Server message are intentionally not trabslated because server seems English only but I can do them too.
- When everithing is ready I'll rebase on master to avoid weird looking commit tree


reinder:
> Before merging I'd like to remove old translations and add missing ones.
That would be nice :)

> Fixed a typo in English translation (can be applied from different PR if you prefer)
This PR is fine :)

> Other language need to be update to show Italian language and also installer needs to be update too.
Yes, please add them :)

> Server message are intentionally not trabslated because server seems English only but I can do them too.
Messages are also display on the client if you open the server log (F12), if they are translated they will be visible in the correct language :) (only the log message code and arguments are sent over the network)

> When everithing is ready I'll rebase on master to avoid weird looking commit tree
👍

gfgit:
Hi, I've completed the tasks so I think it's ready to merge.
Please check  Nederlands translation for the added Italian language name and also `interface.hsi88:modules_middle` which seems to have same typo as English translation

reinder:
Thanks, see:
https://www.facebook.com/112299137122975/posts/727857598900456/?flite=scwspnss

PR #30, opened 2023-02-06
Add MinGW section to BUILDING.md

gfgit:
Added instructions specific to MinGW.
Linked GDB online documentation.

I must say I don't like the document layout because instructions are duplicated almost same for Client and Server.
But that's a minor thing.

I added information about debugging which are entirely missing for other setups.
Maybe something should be added for them too, and something is common to many compilers


reinder:
> I must say I don't like the document layout because instructions are duplicated almost same for Client and Server.
But that's a minor thing.

True, there is some duplication, but also some minor differences. It more matrix like: OS vs compiler vs program.
I choose: program -> OS -> compiler for now.

Maybe OS first, then compiler, then program is better, what do you suggest?

PR #31, opened 2023-02-06
model: HACK to let client refresh after object is removed

gfgit:
When an object is removed from a list, Clients need to clear cached values and reload the model.
This could be more efficient but it will be adressed in a separate pull request.

We can open a new issue specific to Client side sorting and caching


reinder:
> We can open a new issue specific to Client side sorting and caching

See #32 :)

Issue #35, opened 2023-02-13
Log Invalid parameter passed to C runtime


gfgit:
**Describe the bug**
Hi, I'm getting a lot of "Invalid parameter passed to C runtime" running server built with MinGW 12.2.0.

I suspect the warning message is sent after calling `std::put_time()` and it's `operator<<` from [CosoleLogger](https://github.com/traintastic/traintastic/blob/7397cb9e16652d880abf63079155a7a4b34f179b/server/src/log/consolelogger.cpp#L48)

The message is sent early and sometimes gets printed before other calls to `std::cout` or `printf()` so I first thought it was `localtime_s()` fault but now I know it's not that.

The variables seem fine and also passed format argument `%F %T` but it doesn't print anything

gfgit:
[CppReference](https://en.cppreference.com/w/cpp/io/manip/put_time) says that "T" and "F" are available only C++11.
"%H:%M:%S" works so maybe the C runtime is not C++11?

reinder:
C++11 means since C++11 on CppReference.

It seems to be a MinGW bug, see https://sourceforge.net/p/mingw-w64/bugs/793/

Think it is best to add a #define with the log format, then we can define it differently for MinGW and add a comment about the MinGW bug. The [FileLogger](https://github.com/traintastic/traintastic/blob/7397cb9e16652d880abf63079155a7a4b34f179b/server/src/log/filelogger.cpp#L59) also uses them using a semicolon instead of a space as separator.

PR #36, opened 2023-02-14
Logger workaround for MinGW broken std::put_time()

gfgit:
- Conform ConsoleLogger and FileLogger to use same format in 'std::put_time()'

- Workaround missing specifier support in MinGW 12.2.0 by using old pre-C+ +11 specifiers.

See #35

### DIFF
--- a/server/src/log/consolelogger.cpp
+++ b/server/src/log/consolelogger.cpp
@@ -41,11 +41,13 @@ void ConsoleLogger::write(const std::chrono::system_clock::time_point& time, std
   const auto us = std::chrono::duration_cast<std::chrono::microseconds>(time.time_since_epoch()) % 1000000;
   tm tm;
 
+  const char time_format[] = TRAINTASTIC_LOG_DATE_FORMAT " " TRAINTASTIC_LOG_TIME_FORMAT;
+
   std::lock_guard<std::mutex> lock(m_streamMutex);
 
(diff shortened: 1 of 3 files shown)


reinder:
There is a difference in the log format, the file logger has a `;` as seperator, the console logger a ` `.

I'd suggest to do `#define LOG_FORMAT_DATE`  and `#define LOG_FORMAT_TIME` then:

console logger: `LOG_FORMAT_DATE " " LOG_FORMAT_TIME`
file logger: `LOG_FORMAT_DATE ";" LOG_FORMAT_TIME`

if performance wise `%F` is faster than `%Y-%m-%d`, then we should only use `%Y-%m-%d`, for MinGW

gfgit:
> There is a difference in the log format, the file logger has a `;` as seperator, the console logger a ` `.

Sorry, I thought it was a mistake to fix.

I don't know about performance difference, I think it depends on standard library implementation. I don't have any way to test it at the moment.

gfgit:
I've pushed a new version. I was tempted to use `constexpr` string concatenation but is too complex for the task.
Definition are prefixed with `TRAINTASTIC_` because I think `LOG...` has many chance to conflict in the future with other libraries definitions

reinder:
> Definition are prefixed with `TRAINTASTIC_` because I think `LOG...` has many chance to conflict in the future with other libraries definitions

Yes, indeed, I agree :)

Pipeline is running, if it is successful, I'll merge it :)

PR #37, opened 2023-02-26
Add 4 new positions to Double Slip Switch turnout

### LINKED ISSUE #15: Double slip Turnout positions
Hi, I just noticed double slip Switch have "Crossed" and "Diverged" position. Like this I think: Image

But also this Type exists, which support only one route at a time:
Image 2

gfgit:
- This adds Left/Right to double switch
- And adds new DoubleSlipStraightA / DoubleSlipStraightB to TurnoutPosition enum
- Implemented new state drawing in TilePainter
- Updated logic in SignalPath::findBlocks()

Hi, this PR implements features of #15.
It's a quick test so might break code style but the logic seems fine.
Maybe the names of new states should be changed...
Also it would be good to initialize `outputMap.setValueInternal()` with `positionValues` array to avoid duplicating the list.

Closes #15

### DIFF
--- a/client/src/board/tilepainter.cpp
+++ b/client/src/board/tilepainter.cpp
@@ -486,6 +486,14 @@ void TilePainter::drawTurnout(TileId id, const QRectF& r, TileRotate rotate, Tur
       setTurnoutStatePen();
       switch(position)
       {
+        case TurnoutPosition::Left:
+          drawCurve45(turnoutStateRect(r), rotate);
+          break;
+
+        case TurnoutPosition::Right:
+          drawCurve45(turnoutStateRect(r), rotate + TileRotate::Deg180);
+          break;
+
         case TurnoutPosition::Crossed:
           drawStraight(turnoutStateRect(r), rotate);
           drawStraight(turnoutStateRect(r), rotate - TileRotate::Deg45);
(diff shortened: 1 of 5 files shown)


reinder:
Looking good. Yeah naming the enum values is the hardest part I think.
Maybe we can name the enum's just StraightA and StraightB, then they can be reused for single slip, because Single slip should be adjusted too.

The duplicate array/lists is indeed something that should be removed. I think `tcb::span` (= `std::span` for C++17) can be used here also.

Now that there are multiple options for control, the *use* column should be implemented too (or an option to specify how it should be controlled)

gfgit:
> because Single slip should be adjusted too

Single slips actually allow both ways like a normal  diamond crossing so the Straight / Diverged pair is correct already

reinder:
It mainly depends on how the single/double slip is controlled. The Fleischmann double slips I have just require a single motor. I just checked the peco website, their single slip seems to require two motors. See [here](https://peco-uk.com/products/crossing-single-slip4)

If two motors are used it can still be controlled using *crossed* and *diverged*, but then it may drive a motor while it isn't strictly nessacary for a route.

Think it is best for both that the user can choose how to control based on the number of moters than can be controlled individually.

PR #41, opened 2023-03-25
Fix train vehicle list

gfgit:
This fixes move Up/Down not disabled when train is moving.

In the second commit I've fixed a clang warning not strictly related to this issue, maybe there are more of them in other places.


gfgit:
Enabling edit mode while train is moving still enables Name field

PR #49, opened 2023-04-10
Add more Z21 types

gfgit:
Taken from Z21 LAN Protocol Specification 1.12 en


gfgit:
I'm testing an emulated Z21 based on Philipp Gahtow work for Arduino. [See here](https://github.com/gfgit/Z21ServerEmulator)
It is not working properly so I will try to fix some of the issues

reinder:
Mapping short circuit to power off seems right, there won't be power on the track anymore.

I'll merge it.

PR #51, opened 2023-04-11
server: Z21 fix bradcast flags and other small fixes

gfgit:
Request BC flags as keep alive timer until we get a valid reply

Maybe we should allow a Z21 not supporting BC flags? Currently if Z21 replies "None", traintastic keeps asking again.
This can be because:
- the "set" part was skipped and we received only original BC flags -> ask again (current behaviour)
- The Z21 does not support BC flags so "None" is actually the correct status -> switch to normal keep alive message

I think we should consider also the second case

### DIFF
--- a/server/src/hardware/protocol/z21/clientkernel.cpp
+++ b/server/src/hardware/protocol/z21/clientkernel.cpp
@@ -220,11 +220,17 @@ void ClientKernel::receive(const Message& message)
         }
       }
       break;
+    case LAN_GET_BROADCASTFLAGS:
+      if(message.dataLen() == sizeof(LanGetBroadcastFlagsReply))
+      {
+        const auto& reply = static_cast<const LanGetBroadcastFlagsReply&>(message);
+        m_broadcastFlags = reply.broadcastFlags();
+      }
+      break;
 
     case LAN_GET_CODE:
     case LAN_LOGOFF:
     case LAN_SET_BROADCASTFLAGS:
-    case LAN_GET_BROADCASTFLAGS:
     case LAN_GET_LOCO_MODE:
     case LAN_SET_LOCO_MODE:
     case LAN_GET_TURNOUTMODE:
@@ -435,6 +441,7 @@ void ClientKernel::simulateInputChange(uint32_t channel, uint32_t address, Simul
 void ClientKernel::onStart()
 {
   // reset all state values
+  m_broadcastFlags = BroadcastFlags::None;
   m_serialNumber = 0;
   m_hardwareType = HWT_UNKNOWN;
   m_firmwareVersionMajor = 0;
@@ -447,12 +454,7 @@ void ClientKernel::onStart()
   send(LanGetSerialNumber());
   send(LanGetHardwareInfo());
 
-  send(LanSetBroadcastFlags(
-    BroadcastFlags::PowerLocoTurnoutChanges |
-    BroadcastFlags::RBusChanges |
-    BroadcastFlags::SystemStatusChanges |
-    BroadcastFlags::AllLocoChanges | // seems not to work with DR5000
-    BroadcastFlags::LocoNetDetector));
+  send(LanSetBroadcastFlags(requiredBroadcastFlags));
 
   send(LanGetBroadcastFlags());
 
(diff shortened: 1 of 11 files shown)


server/src/hardware/protocol/z21/clientkernel.cpp
```diff
@@ -221,17 +221,61 @@ void ClientKernel::receive(const Message& message)
...
+    {
+      if(message.dataLen() == sizeof(LanSystemStateDataChanged))
+      {
+        const auto& reply = static_cast<const LanSystemStateDataChanged&>(message);
+
+        const bool isStop = reply.centralState & Z21_CENTRALSTATE_EMERGENCYSTOP;
+        const bool isTrackPowerOn = (reply.centralState & Z21_CENTRALSTATE_TRACKVOLTAGEOFF) == 0
+                                    && (reply.centralState & Z21_CENTRALSTATE_SHORTCIRCUIT) == 0;
+
+        const TriState trackPowerOn = isTrackPowerOn ? TriState::True : TriState::False;
```
reinder, line 242: Please use `constexpr TriState toTriState(bool value)`

server/src/hardware/protocol/z21/clientkernel.hpp
```diff
@@ -62,6 +62,14 @@ class ClientKernel final : public Kernel
   private:
     const bool m_simulation;
     boost::asio::steady_timer m_keepAliveTimer;
+    BroadcastFlags m_broadcastFlags;
+
+    static constexpr BroadcastFlags RequiredBCFlags =
```
reinder, line 67: Please rename it to `requiredBroadcastFlags`.
Variable names are lowerCamelCase.
I prefer to don't abbreviate, just to be clear.

server/src/hardware/protocol/z21/clientkernel.cpp
```diff
@@ -483,8 +523,18 @@ void ClientKernel::send(const Message& message)
 
 void ClientKernel::startKeepAliveTimer()
 {
-  assert(ClientConfig::keepAliveInterval > 0);
-  m_keepAliveTimer.expires_after(boost::asio::chrono::seconds(ClientConfig::keepAliveInterval));
+  if(m_broadcastFlags == BroadcastFlags::None)
```
reinder, line 526: Shoudn't we test against the `requiredBroadcastFlags` ?

reinder (commented):
We could add a retry count for setting the broadcast flags, e.g. try 10 times if it still isn't correct after 10 retries log a warning and continue with keep alive.

server/src/hardware/protocol/z21/clientkernel.cpp
gfgit: > Shoudn't we test against the `requiredBroadcastFlags` ?

For the arduino Z21 implementation, some flags are not supported (e.g. CAN Bus) and some other are forced to be enabled together (e.g. enabling something about LocoNet, it enables every LocoNet flag).
So in general there is a negotiation on the flag value, it may be more than what was requested or some features may be missing.

reinder: I think it would be good to log a warning if there is a mismatch or the retry count expires.
Then the user at least knows something might not work as expected.

You can add two warnings for that in the 2xxx range.

gfgit: For now I've added only one message...

Two things:
- Do you think it's important also to log the effective returned flags?
- Also do we consider a mismatch if reply has more flags than requested?
Technically it's still a mismatch, but we receive all things requested and just ignore extra broadcasted messages we are not interested in.
We can test it like: `(flags & requiredFlags) == requiredFlags`

reinder: > Do you think it's important also to log the effective returned flags?

I'm a bit in doubt about it, for us as developers it it useful information, however for an average user it is useless. If e.g. someone in the future runs into it, we can just instruct them to enable debug logging and send the log. That will give us as developers much more information to work with. So my vote is for don't log the flags in the message. What is your opinion?

> Also do we consider a mismatch if reply has more flags than requested?
Technically it's still a mismatch, but we receive all things requested and just ignore extra broadcasted messages we are not interested in.

Good point, It is still a mismatch, Traintastic can behave as expected so no need to bother the user with it. We could how log it as extra debug message if debug logging is enabled (makes it a bit easier to spot for us as developers).
What is your opinion?

gfgit: I agree with you. So we will have 2 messages:
- A warning emitted only if BC flags are less than requested
- A debug (hidden by default) emitted always and with flag value

reinder: Yes indeed :)

gfgit:
I've split turnout related changes in PR #56, so they can be further tested
The remaining changes are small fixes so they can be already merged

PR #54, opened 2023-04-25
Sync Train speed from rail vehicles

gfgit:
Follows #44


gfgit:
@reinder
Hi, I'm trying to implement vehicle-specific and train-specific settings for trains, I'm taking `BlockInputMapItem` as an example.
This should be used to:
- Invert single rail vehicle direction in case it's coupled reversed
When a vehicle is run through a reverse loop, or manually swapped by hand, it's direction is opposite to other vehicles
Take this into account when setting train direction so every locomotive moves correctly
- Set powered vehicle in idle to simulate it being powered off and moved by another locomotive
This means sound effects are turned off by default and do not react to global train functions (behave like a wagon)
Maybe also turn off light by default
But model train motor is running to keep same speed of other locomotives in the train (behave like a locomotive)
```
class TrainVehicleListItem : public Object
{
public:
    CLASS_ID("list.train_vehicle.item")

    ObjectProperty<RailVehicle> vehicle;
    Property<bool> invertDirection;
    Property<bool> idle;
};
```

I'm a bit confused about how to make it work, if I need to override `getObjectId()`, change `TrainVehicleList` class etc...

gfgit:
[b5cf512](https://github.com/traintastic/traintastic/pull/54/commits/b5cf512f0d2d6d9cfd2c489ac6b40077143268ca) is a test to forward declare `TrainVehicleList`.
I get strange `std::dynamic_pointer_cast` errors if I omit the inclusion in the modified files.
It seems that `objectproperty.tpp` is implementing `toObject()` also for `Train::vehicles` even if it's not called by those files directly...
So it fails because `TrainVehicleList` is incomplete type without explicit include.

gfgit:
[5123d74](https://github.com/traintastic/traintastic/pull/54/commits/5123d74f6539e0b8fd96ee3772375dea3ce7e66e) is not working because of assert on `TrainVehicleList` which must subclass either `SubObject` or `IdObject`.
Both do not allow to override `getObjectId()` because is marked final...
If I subclass from `IdObject` I have issues having both item id and vehicle id mixed up

reinder:
> @reinder Hi, I'm trying to implement vehicle-specific and train-specific settings for trains, I'm taking `BlockInputMapItem` as an example. This should be used to:
> 
>     * Invert single rail vehicle direction in case it's coupled reversed
>       When a vehicle is run through a reverse loop, or manually swapped by hand, it's direction is opposite to other vehicles
>       Take this into account when setting train direction so every locomotive moves correctly
> 
>     * Set powered vehicle in idle to simulate it being powered off and moved by another locomotive
>       This means sound effects are turned off by default and do not react to global train functions (behave like a wagon)
>       Maybe also turn off light by default
>       But model train motor is running to keep same speed of other locomotives in the train (behave like a locomotive)

Yeah that sounds good, we need to associate additional properties to vehicles in a specific train. It could be done how you propose or by adding a seperate list/vector for storing the extra properties, but I agree on the one you purpose :)
 
> ```
> class TrainVehicleListItem : public Object
> {
> public:
>     CLASS_ID("list.train_vehicle.item")
> 
>     ObjectProperty<RailVehicle> vehicle;
>     Property<bool> invertDirection;
>     Property<bool> idle;
> };
> ```
> 
> I'm a bit confused about how to make it work, if I need to override `getObjectId()`, change `TrainVehicleList` class etc...

`getObjectId()` is required to get a unique id for the network communication and JSON save/restore and can be used for the `world.get_object()` function in Lua. You can just generate a unique id using the `World::getUniqueId()` method.

> [5123d74](https://github.com/traintastic/traintastic/pull/54/commits/5123d74f6539e0b8fd96ee3772375dea3ce7e66e) is not working because of assert on `TrainVehicleList` which must subclass either `SubObject` or `IdObject`. Both do not allow to override `getObjectId()` because is marked final... If I subclass from `IdObject` I have issues having both item id and vehicle id mixed up

Just change the `ObjectList` to require `Object` instead of `IdObject` || `SubObject`, it is just there because that was true, it was `ObjectId` at first, later I added `SubObject` because that was required to. We will see if we run into other issues :)

p.s. Don't know what is *best practice* regarding copyright, but for files you create from scratch I think it makes sense that you put your name there. Also for files you make a great contribution to it makes sense to add an extra copyright line.
p.s.2. You can run the `utils/check-sources.py` script, that will correct the include guard and add a GPLv2 header (with my name...I need to change that to read it from git :))
p.s.3. I saw at another project they have: Copyright $year $main_author and contributors, seem fair to me :)

reinder:
> [5123d74](https://github.com/traintastic/traintastic/pull/54/commits/5123d74f6539e0b8fd96ee3772375dea3ce7e66e) is not working because of assert on `TrainVehicleList` which must subclass either `SubObject` or `IdObject`. Both do not allow to override `getObjectId()` because is marked final... If I subclass from `IdObject` I have issues having both item id and vehicle id mixed up

Yeah I ran into this to, needing to add includes in other files even if it isn't used. Didn't find a solution yet...

gfgit:
New version of train-specific settings: now `ObjectVectorProperty` is used instead of `ObjectList`
Previous strategy did not work because:

- `ObjectList` wants items to subclass `SubObject` which is not what we want here
- `getObjectByPath()` failed to get items of `ObjectList` as sub elements of parent Train

**NOTE:** There is a bug:

1. Add a new Train in a new World
2. Add a vehicle to it
3. Remove the vehicle
4. Train is now empty again
5. Add again a vehicle
6. Train shows one line for the vehicle bug is empty

Subsequent added vehicles show up correctly but first row remains empty.
Temp workaround: close/reopen train or reverse list 2 times to trigger full list refresh

TODO: make items editable only when Train is not running like other properties

gfgit:
- Fixed loading of `TrainVehicleList` now properly using "state"
- Fixed powered rail vehicles detection
- New experimental handling of decoder direction change -> update whole train direction

Some thought on last change:
- It triggers segmentation fault so I cannot test it properly yet, now I will compile with `-ggdb3` hoping to get more debug info
- Should direction change regard only PoweredRailVehicles or all contained vehicles? (as even non-powered vehicles can be assigned a decoder, hence have a direction)
- Should we assert vehicle is contained in train?
- Is same direction value check redundant?
- How to handle direction unknown properly?

Still have to fix previous message issues...

gfgit:
Rebased on current master
- Fixed segmentation fault which was caused by infinite recursion in decoder changed handler
- Sync all vehicles direction when train becomes active
- Disable vehicles items when train is running
- Fixed train direction propagation when not active

gfgit:
Now I get a strange compile error:
```
server/src/network/session.cpp:66:28: error: 'GetObject' is not a member of 'Message::Command'; did you mean 'GetObjectA'?
   66 |     case Message::Command::GetObject:
      |                            ^~~~~~~~~
      |                            GetObjectA
```

This is because `GetObject` is also a macro defined in `<wingdi.h>` but it always compiled fine until now...
Don't know how recent changes are related to this.

gfgit:
Maybe we should split all templeated classes implementations in `.tpp` files.
`Method` class also complains about incomplete type

gfgit:
There is now a `fixup!` commit which fixes headers to make it compile.
It adds `BlockRailTile` to `TrainVehicleList` which should not be needed. I'm keeping it separate hoping to remove it when we sort out this forward declaration problem.

reinder:
Yes, indeed, when `ObjectProperty` was split into `hpp` and `tpp`  it also required additional includes which shouldn't be needed as it isn't used in those files.

server/src/train/trainvehiclelist.cpp
```diff
@@ -89,27 +94,137 @@ TrainVehicleList::TrainVehicleList(Train& train_, std::string_view parentPropert
   m_interfaceItems.add(reverse);
 }
 
+void TrainVehicleList::load(WorldLoader& loader, const nlohmann::json& data)
+{
+  nlohmann::json state = loader.getState(getObjectId());
+  nlohmann::json objects = state.value("items", nlohmann::json::array());
+  if(!objects.empty())
+  {
+      std::vector<std::shared_ptr<TrainVehicleListItem>> values;
```
reinder, line 103: indent not 2 spaces

server/src/train/trainvehiclelist.cpp
```diff
@@ -89,27 +94,137 @@ TrainVehicleList::TrainVehicleList(Train& train_, std::string_view parentPropert
...
+  }
+}
+
+void TrainVehicleList::addObject(std::shared_ptr<RailVehicle> vehicle)
+{
+  std::shared_ptr<TrainVehicleListItem> object = getItemFromVehicle(vehicle);
+  if(object)
+    return; //Vehicle is already in this train
+
+  object = std::make_shared<TrainVehicleListItem>(*this, getItemId());
```
reinder, line 176: A `TrainVehicleListItem` should always have a vehicle, why not add it to the constructor to make it more clear that it is required

server/src/train/trainvehiclelist.cpp
```diff
@@ -89,27 +94,137 @@ TrainVehicleList::TrainVehicleList(Train& train_, std::string_view parentPropert
...
+
+  train().vehiclesChanged();
 }
 
 Train& TrainVehicleList::train()
 {
   return static_cast<Train&>(parent());
 }
+
+unsigned int TrainVehicleList::getItemId() const
```
reinder, line 214: unsigned int -> uint32_t

just as in the function (they are the same usally, but it look weird)

server/src/train/trainvehiclelist.cpp
```diff
@@ -89,27 +94,137 @@ TrainVehicleList::TrainVehicleList(Train& train_, std::string_view parentPropert
...
+{
+  uint32_t itemId = 1;
+  while(std::find_if(begin(), end(), [itemId](const auto& it){ return it->itemId() == itemId; }) != end())
+      itemId++;
+  return itemId;
+}
+
+std::shared_ptr<TrainVehicleListItem> TrainVehicleList::getItemFromVehicle(const std::shared_ptr<RailVehicle> vehicle)
```
reinder, line 222: const ref shared_ptr  for argument and return type

for a null return add: `static const std::shared_ptr<TrainVehicleListItem> null;` in the function, the `null` can be returned const ref.

server/src/train/trainvehiclelist.cpp
```diff
@@ -89,27 +94,137 @@ TrainVehicleList::TrainVehicleList(Train& train_, std::string_view parentPropert
...
+{
+  auto it = std::find_if(begin(), end(), [vehicle](std::shared_ptr<TrainVehicleListItem> item)
```
reinder, line 224: lambda argument should be const ref

Please put the whole lambda on a new line (just like almost all other lambda's)

server/src/train/trainvehiclelistitem.cpp
```diff
@@ -0,0 +1,111 @@
+/**
+ * server/src/train/trainvehiclelistitem.cpp
+ *
+ * This file is part of the traintastic source code.
+ *
+ * Copyright (C) 2023 Reinder Feenstra
```
reinder, line 6: I think you should put your name here ;)

server/src/train/trainvehiclelistitem.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+
+TrainVehicleListItem::TrainVehicleListItem(TrainVehicleList &parent, uint32_t itemId) :
+  m_parent(parent),
+  m_itemId(itemId),
+  vehicle(this, "vehicle", nullptr, PropertyFlags::ReadWrite | PropertyFlags::ScriptReadWrite | PropertyFlags::StoreState,
+    [/*this*/](const std::shared_ptr<RailVehicle>& /*value*/)
+    {
+      //propertyChanged(vehicle);
+    },
```
reinder, line 38: replace by `nullptr` ?

server/src/train/trainvehiclelistitem.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include "trainvehiclelistitem.hpp"
+#include "trainvehiclelist.hpp"
+#include "../world/getworld.hpp"
+#include "../core/attributes.hpp"
+#include "../core/objectproperty.tpp"
+#include "../vehicle/rail/railvehicle.hpp"
+
+
+TrainVehicleListItem::TrainVehicleListItem(TrainVehicleList &parent, uint32_t itemId) :
```
reinder, line 31: I prefere to use the style:
```
Object::Object()
  : Something()
  , m_something
{
```

Because it results in less changes if things are added, you now don't need too add the trailing comma on the previous line :)

server/src/train/trainvehiclelistitem.hpp
```diff
@@ -0,0 +1,62 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+#ifndef TRAINVEHICLELISTITEM_H
```
reinder, line 23: please run `utils/check-sources.py` to correct the include guard

server/src/train/trainvehiclelisttablemodel.cpp
```diff
@@ -0,0 +1,115 @@
...
+            return item.vehicle->name;
+
+        case columnType:
+            return std::string("$class_id:").append(item.vehicle->getClassId()).append("$");
+
+        case columnLOB:
+            return toString(item.vehicle->lob);
+
+        case columnInvertDir:
+            return item.invertDirection ? "invert" : "normal";
```
reinder, line 94: These need translation, other option is a checkmark if inverted, else nothing.

server/src/train/trainvehiclelistitem.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+    void connectVehicle(RailVehicle &object);
+    void disconnectVehicle(RailVehicle &);
+
+  public:
+    ObjectProperty<RailVehicle> vehicle;
+    Property<bool> invertDirection;
+
+    TrainVehicleListItem(TrainVehicleList& parent, uint32_t itemId);
+    ~TrainVehicleListItem();
```
reinder, line 56: missing override

reinder (changes requested):
I'm not yet a big fan of limiting the throttle value based on the decoder speed steps. Scaling differs per protocol. I think it is best to let the protocol kernel handle it.

Before merging it I'd like to do some tests with real hardware and see if it works as expected. I can test LocoNet/XpressNet/Z21 using a Digikeijs DR5000.

reinder:
For direction invert we could overload the `!` operator for the `Direction` enum.

reinder:
I was just thinking, maybe we should remove the float throttle property and replace it by speedstep. The interface should then control the decoders protocol/speedsteps options as that is different per interface and protocol.

Mapping km/h to a speedstep for the speed profile may also be easier than.

Just another option, we need to find out what works best.

server/src/train/train.hpp
```diff
@@ -61,6 +65,8 @@ class Train : public IdObject
     void updateSpeedMax();
     void updateEnabled();
     bool setTrainActive(bool val);
+    void propagateDirection(Direction newDirection);
+    void handleDecoderDirection(std::shared_ptr<PoweredRailVehicle> vehicle, Direction newDirection);
```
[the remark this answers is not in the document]
gfgit: I think ref defeats the purpose of shared ptr but maybe in this case it's ok

server/src/train/trainvehiclelist.cpp
gfgit: Should vehicle property be readonly then?
Like to change vehicle you delete and readd the item.

server/src/train/trainvehiclelistitem.cpp
gfgit: It's not consistently used in other parts of code. We really should clang format everithing to keep consistency

gfgit:
> Mapping km/h to a speedstep for the speed profile may also be easier than.

Ok but then when you change protocol/interfaces for a decoder you will need to rework the saved speed profile.
But I agree float is not ideali either

gfgit:
Would you prefer new commits for the corrections or a rebase to keep everithing tidy?

server/src/board/board.cpp
```diff
@@ -20,6 +20,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  */
 
+#include "../train/trainvehiclelist.hpp"
```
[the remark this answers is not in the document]
gfgit: Maybe I found the issue.
The `CREATE` macro instantiates an instance in the header where some classes are still forward declared.
It shouldn't be a problem but we can try splitting in `CREATE` in header and `CREATE_IMPL` in source file

server/src/train/trainvehiclelist.cpp
gfgit: Then BlockInputMap has the same issue and maybe also other classes

server/src/train/trainvehiclelistitem.hpp
gfgit:
```
Traceback (most recent call last):
  File "C:\Filippo\Qt_Project\traintastic\utils\check-sources.py", line 102, in <module>
    check_sources(os.path.abspath(os.path.join(PROJECT_ROOT, source_dir)), fix)
  File "C:\Filippo\Qt_Project\traintastic\utils\check-sources.py", line 78, in check_sources
    text = text_org = f.read()
                      ^^^^^^^^
  File "C:\Users\Filippo\AppData\Local\Programs\Python\Python311\Lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 5991: character maps to <undefined>
```

gfgit: It needs `open(..., encoding='utf8')` I think

server/src/train/trainvehiclelist.cpp
reinder: Yes, I think that is better. As long as we don't add a *replace vehicle* option it is not required and event then we will problably use `setValueInteral`.

server/src/board/board.cpp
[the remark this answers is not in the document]
reinder: I already did....two months ago 🙃, see [objectid.hpp](https://github.com/traintastic/traintastic/commit/246b27c4344f24e4be101e79e72db0a168f53dad#diff-2f09990810b0b933d2387516361f41ab0f68f0e8b2411a1beacd425488bd8dcc), maybe we should do that for all `CREATE`'s

server/src/train/trainvehiclelistitem.cpp
reinder: Older parts of the code base use the older style, clang-format would be best indeed, I did some experiments with it, but couldn't get it setup so it doesn't cause to much changes.

server/src/train/trainvehiclelistitem.hpp
reinder: On linux it works fine, but not for windows, please add it :)

Just to be sure you are using Python 3?

server/src/train/trainvehiclelist.cpp
reinder: You're right there is something wrong there, my bad:
```
  private:
    unsigned int getItemId() const;
```

```
uint32_t BlockInputMap::getItemId() const
{
  uint32_t itemId = 1;
  while(std::find_if(items.begin(), items.end(), [itemId](const auto& it){ return it->itemId() == itemId; }) != items.end())
    itemId++;
  return itemId;
}
```

I'll correct that to `uint32_t`

reinder:
> Ok but then when you change protocol/interfaces for a decoder you will need to rework the saved speed profile.
But I agree float is not ideali either

True, we can try to convert the speed profile to the new protocol/speed steps as good as possible.
Changing things after calibration can always affect the speed profile, luckly most (future) users have just one command station :)

We shoud really add some export/import support for speed profiles (and trains etc. then you can easily transfer that data to another world, or e.g. a club layout)

server/src/train/trainvehiclelistitem.hpp
gfgit: Yes 3.11

gfgit:
I will not rebase because it would i validate revuews (it changes commit ID)

server/src/train/trainvehiclelist.cpp
gfgit: Already done in this PR

gfgit:
@reinder 
There is a problem when adding an high speed locomotive (say max speed 300 km/h) to a train.
When train starts accelerating from 0, first speed will be 0.36 km/h which due to rounding to 128 steps gets rounded to zero.
This then breaks the logic because float throttle is not zero but received throttle is exactly zero.
This is also not good because when starting, hoping to reach higher speed (0.36 km/h is just an intermediate state), all vehicles
should start moving, otherwise we get tension on couplings

gfgit:
Added check to prevent exceeding train maximum speed when operating from handset controller or any other "single decoder" controllers

gfgit:
> I will not rebase because it would invalidate revuews (it changes commit ID)

This PR is become too messy so I did a big rebase in the end :)

gfgit:
Rebased on #64

reinder:
I'll try to give this one a test drive soon, looking forward to it!

gfgit:
Cannot build it anymore:
```
ld.lld: error: undefined symbol: TrainVehicleListTableModel::isListedProperty(std::basic_string_view<char, std::char_traits<char>>)
>>> referenced by C:/Qt_Project/traintastic/server/src/train\trainvehiclelist.cpp:129
>>>               CMakeFiles/traintastic-server.dir/src/train/trainvehiclelist.cpp.obj:(TrainVehicleList::propertyChanged(BaseProperty&))

ld.lld: error: undefined symbol: TrainVehicleListTableModel::propertyChanged(BaseProperty&, unsigned int)
>>> referenced by C:/Qt_Project/traintastic/server/src/train\trainvehiclelist.cpp:139
>>>               CMakeFiles/traintastic-server.dir/src/train/trainvehiclelist.cpp.obj:(TrainVehicleList::propertyChanged(BaseProperty&))

ld.lld: error: undefined symbol: TrainVehicleListItem::TrainVehicleListItem(TrainVehicleList&, unsigned int)
>>> referenced by C:/bin/mingw64/lib/gcc/x86_64-w64-mingw32/12.2.0/include/c++/bits\stl_construct.h:119
>>>               CMakeFiles/traintastic-server.dir/src/train/trainvehiclelist.cpp.obj:(void std::_Construct<TrainVehicleListItem, TrainVehicleList&, unsigned int>(TrainVehicleListItem*, TrainVehicleList&, unsigned int&&))

ld.lld: error: undefined symbol: TrainVehicleListTableModel::TrainVehicleListTableModel(TrainVehicleList&)
>>> referenced by C:/bin/mingw64/lib/gcc/x86_64-w64-mingw32/12.2.0/include/c++/bits\stl_construct.h:119
>>>               CMakeFiles/traintastic-server.dir/src/train/trainvehiclelist.cpp.obj:(void std::_Construct<TrainVehicleListTableModel, TrainVehicleList&>(TrainVehicleListTableModel*, TrainVehicleList&))
collect2.exe: error: ld returned 1 exit status
```

Issue #21, opened 2023-01-23
Simplify turnout/signal output mapping


reinder:
Currently output mapping is a lot of work, the user needs to:
- Create outputs on the interface
- Then assign these to the turnout/signal
- Configure them

For most common situations it is much easier to just assign the interface and turnout/signal address, the software should create the output mapping automatically. The *advanced mode* should still be possible in case that is needed.

reinder:
TODO: Support DDCext too, e.g. the [YaMoRC YD8116](http://yamorc.de/products/?singleproduct=1247) supports that.

gfgit:
Should we support external turnout switching? Like if I switch a turnout from command station should the corresponding `Output` and `TurnoutRailTile` update their state?
Currently it doesn't seem to work

reinder:
Yes, that is something that must be added.

This also raises some issues, if a turnout is locked (e.g. in the future by a reserved train path). How should that be handled?
- Send a command to set the turnout state back
- Estop the train and remove the reserved path
- ...

gfgit:
Also turnout set to "pulse" output.
How do you detect incoming pulse?

reinder:
A transition from low "0" to high "1", the strart of a pulse should be fine I think. It isn't a complete pulse, but that will problably work with most systems. Some systems also handle the pulse in hardware, when you set a output high the command stations sets it low after ~250ms or so. It is a bit gray area, that will require some testing to find "best practice" and what the user expects.

PR #56, opened 2023-04-25
Z21 fix turnout handling

gfgit:
Needs #81 applied first. Related to #65 

- Fix turnout addressing
- Get turnout info back

### DIFF
--- a/server/src/hardware/protocol/z21/clientkernel.cpp
+++ b/server/src/hardware/protocol/z21/clientkernel.cpp
@@ -26,6 +26,7 @@
 #include "../../decoder/decoderchangeflags.hpp"
 #include "../../protocol/dcc/dcc.hpp"
 #include "../../input/inputcontroller.hpp"
+#include "../../output/outputcontroller.hpp"
 #include "../../../core/eventloop.hpp"
 #include "../../../log/log.hpp"
 #include "../../../utils/inrange.hpp"
@@ -70,6 +71,24 @@ void ClientKernel::receive(const Message& message)
 
       switch(lanX.xheader)
       {
+        case LAN_X_TURNOUT_INFO:
+          if(message.dataLen() == sizeof(LanXTurnoutInfo))
+          {
+            const auto& reply = static_cast<const LanXTurnoutInfo&>(message);
+            OutputPairValue value = OutputPairValue::Undefined;
+            if(!reply.positionUnknown())
+            {
+              value = reply.state() ? OutputPairValue::Second : OutputPairValue::First;
+            }
+
+            EventLoop::call(
+              [this, address=reply.address(), value]()
+              {
+                m_outputController->updateOutputValue(OutputChannel::Accessory, address, value);
+              });
+          }
+          break;
+
         case LAN_X_BC:
           if(message == LanXBCTrackPowerOff() || message == LanXBCTrackShortCircuit())
           {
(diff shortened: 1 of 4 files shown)


gfgit:
@reinder
Is this working for you?

reinder:
I'll check it with a DR5000, to verify if the Z21 addresses will match with the LocoNet/XpressNet addresses.

reinder:
I just connected to a DR5000 via LocoNet and Z21 using this branch.

image
I connected a turnout to output 15 and 16, using pair/turnout addressing this is address 8.

image

The address scheme is different. LocoNet has flat addresses (one for every output), the Z21 has them in pairs now (aka "turnout" addresses)

This is a common issue, not all command stations support controlling outputs individually, so Traintastic might need better support for that, also most users are "thinking" in turnout/pair addresses. See also #21.

reinder:
See #65

reinder:
@gfgit is there anything you want to add to this PR? or is it ready?

gfgit:
Do you want to wait for #65?
In the meantime I can extract a subset of changes from my PRs which can be already applied.

gfgit:
Rebased onto #81

reinder:
> Do you want to wait for #65? In the meantime I can extract a subset of changes from my PRs which can be already applied.

I'm a bit in doubt of this one, not because of your work, but because how outputs currently work in Traintastic and how turnouts and signals need to be setup for the end user, see #21.

I think it is wise to do some more investigation (in #65) about the differences between the supported command stations/protocols and how their outputs work and how this affects the setup of turnouts and signals.

reinder:
@gfgit due to #21 and #95 I think it is easier to re-apply these changes to the latest master.

gfgit:
Hi Reinder, I've re-applied some changes on top of master.
I have some questions:

1. The turnout address conversion is now done inside `LanXSetTurnout::setAddress()` and `LanXSetTurnout::address()`. Should same conversion happen for `LanXTurnoutInfo` and `LanXGetTurnoutInfo` ? Z21 specifications seems to treat them in same way.
2. When receiving `LanXTurnoutInfo` message, I call `m_outputController->updateOutputValue(...)` inside `EventLoop`. How do I choose `OutputChannel` enum? Z21 spec seems to allow any accessory address.
3. `OutputPair` seems very similar to `TriState`. Which one should I pass?
4. In some other command stations if value is `TriState::Undefined` then the call to `updateOutputValue` is skipped

reinder:
> 1. The turnout address conversion is now done inside LanXSetTurnout::setAddress() and LanXSetTurnout::address(). Should same conversion happen for LanXTurnoutInfo and LanXGetTurnoutInfo ? Z21 specifications seems to treat them in same way.

Yes please, then when using `address()` or `setAddress()` in you don't have to know how how it is mapped in the message, else every time you read `address()` your have to do the correction between the address in message and the "user" address.

> 2. When receiving `LanXTurnoutInfo` message, I call `m_outputController->updateOutputValue(...)` inside `EventLoop`. How do I choose `OutputChannel` enum? Z21 spec seems to allow any accessory address.

The `OutputChannel` must be one of the options listed in `tcb::span<const OutputChannel> Z21Interface::outputChannels() const`, so in this case  `OutputChannel::Accessory`.

> 3. `OutputPair` seems very similar to `TriState`. Which one should I pass?

`OutputPair`, as the `OutputChannel::Accessory` channel uses output pairs.

> 4. In some other command stations if value is `TriState::Undefined` then the call to `updateOutputValue` is skipped

Yes only if that isn't posible protocol wise.

Hoop this help, if unclear please ask :)

PR #59, opened 2023-05-09
Z21 handle loco info

### LINKED ISSUE #57: Z21 Async nature
Hi, I'm doing some experiments with emulated Z21 moving some decoders.
I've observed infinite async "loops" like this:
```
1. Z21 set speed = 1
2. Z21 set speed = 2
3. Server receives speed 1 (point 1) -> resend "changed" speed = 1
4. Server receives speed 2 (point 2) -> resend "changed" speed = 2
5. Z21 receive speed 1 (from point 3) -> set speed = 1
6. Z21 receive speed 2 (from point 4) -> set speed = 2
7. like 3
8. like 4
... and on ...
```

This happens when rotating the throttle which sends intermediate values until rotation is complete but traintastic is sending "mirrored" values back to Z21 in between.

Possible solution might be remembering "last state set by Z21" and avoid sending "changed" if it's what Z21 already told us

gfgit:
This PR handles receiving loco state from Z21 and update relevant Decoders.

- Needs #64 first
- Fix 28 steps format (not tested but coherent with protocol specifications and JMRI source code)
- Added support for receiving F29 - F31 (When Z21 firmware version is >= 1.42)
- Use helper functions for LanXSetLocoDrive instead of complex to read bit shifting
- Fix infinite loop explained in issue #57 (fixes #57)

### DIFF
--- a/server/src/hardware/protocol/xpressnet/messages.cpp
+++ b/server/src/hardware/protocol/xpressnet/messages.cpp
@@ -25,10 +25,9 @@
 
 namespace XpressNet {
 
-uint8_t calcChecksum(const Message& msg)
+uint8_t calcChecksum(const Message& msg, const int dataSize)
 {
   const uint8_t* p = reinterpret_cast<const uint8_t*>(&msg);
-  const int dataSize = msg.dataSize();
   uint8_t checksum = p[0];
   for(int i = 1; i <= dataSize; i++)
     checksum ^= p[i];
@@ -40,9 +39,9 @@ void updateChecksum(Message& msg)
   *(reinterpret_cast<uint8_t*>(&msg) + msg.dataSize() + 1) = calcChecksum(msg);
 }
 
-bool isChecksumValid(const Message& msg)
+bool isChecksumValid(const Message& msg, const int dataSize)
 {
-  return calcChecksum(msg) == *(reinterpret_cast<const uint8_t*>(&msg) + msg.dataSize() + 1);
+  return calcChecksum(msg, dataSize) == *(reinterpret_cast<const uint8_t*>(&msg) + dataSize + 1);
 }
 
 std::string toString(const Message& message, bool raw)
--- a/server/src/hardware/protocol/xpressnet/messages.hpp
+++ b/server/src/hardware/protocol/xpressnet/messages.hpp
@@ -40,9 +40,14 @@ constexpr uint8_t idFeedbackBroadcast = 0x40;
 
 struct Message;
 
-uint8_t calcChecksum(const Message& msg);
+inline uint8_t calcChecksum(const Message& msg);
+uint8_t calcChecksum(const Message& msg, const int dataSize);
+
 void updateChecksum(Message& msg);
-bool isChecksumValid(const Message& msg);
+
+inline bool isChecksumValid(const Message& msg);
+bool isChecksumValid(const Message& msg, const int dataSize);
+
 std::string toString(const Message& message, bool raw = false);
 
 struct Message
@@ -589,6 +594,16 @@ namespace RoSoftS88XpressNetLI
       checksum = calcChecksum(*this);
     }
   };
+}
+
+inline uint8_t calcChecksum(const Message& msg)
+{
+  return calcChecksum(msg, msg.dataSize());
+}
+
+inline bool isChecksumValid(const Message& msg)
+{
+  return isChecksumValid(msg, msg.dataSize());
 }
 
 }
--- a/server/src/hardware/protocol/z21/clientkernel.cpp
+++ b/server/src/hardware/protocol/z21/clientkernel.cpp
@@ -24,6 +24,7 @@
 #include "messages.hpp"
 #include "../../decoder/decoder.hpp"
 #include "../../decoder/decoderchangeflags.hpp"
+#include "../../protocol/dcc/dcc.hpp"
 #include "../../input/inputcontroller.hpp"
 #include "../../../core/eventloop.hpp"
 #include "../../../log/log.hpp"
@@ -112,6 +113,44 @@ void ClientKernel::receive(const Message& message)
                 });
           }
           break;
+
+        case LAN_X_LOCO_INFO:
+        {
+          const int n = message.dataLen() - 7;
+          if(n >= 7 && n <= 14)
+          {
+            const auto& reply = static_cast<const LanXLocoInfo&>(message);
+
+            bool val[31 + 1] = {};
+            const int maxFunc = (n >= 8) ? 31 : 28;
+            for(int i = 0; i <= maxFunc; i++)
+            {
+              val[i] = reply.getFunction(i);
+            }
+
+            EventLoop::call(
+              [this, address=reply.address(), isEStop=reply.isEmergencyStop(),
+              speed = reply.speedStep(), speedMax=reply.speedSteps(),
+              dir = reply.direction(), val, maxFunc]()
+              {
+                if(auto decoder = m_decoderController->getDecoder(DCC::getProtocol(address), address))
+                {
+                  float throttle = Decoder::speedStepToThrottle(speed, speedMax);
+
+                  decoder->emergencyStop = isEStop;
+                  decoder->direction = dir;
+
+                  decoder->throttle = throttle;
+
+                  for(int i = 0; i <= maxFunc; i++)
+                  {
+                    decoder->setFunctionValue(i, val[i]);
+                  }
+                }
+              });
+          }
+          break;
+        }
       }
       break;
     }
--- a/server/src/hardware/protocol/z21/clientkernel.cpp
+++ b/server/src/hardware/protocol/z21/clientkernel.cpp
@@ -133,19 +133,26 @@ void ClientKernel::receive(const Message& message)
               speed = reply.speedStep(), speedMax=reply.speedSteps(),
               dir = reply.direction(), val, maxFunc]()
               {
-                if(auto decoder = m_decoderController->getDecoder(DCC::getProtocol(address), address))
+                try
                 {
-                  float throttle = Decoder::speedStepToThrottle(speed, speedMax);
-
-                  decoder->emergencyStop = isEStop;
-                  decoder->direction = dir;
-
-                  decoder->throttle = throttle;
-
-                  for(int i = 0; i <= maxFunc; i++)
+                  if(auto decoder = m_decoderController->getDecoder(DCC::getProtocol(address), address))
                   {
-                    decoder->setFunctionValue(i, val[i]);
+                    float throttle = Decoder::speedStepToThrottle(speed, speedMax);
+
+                    decoder->emergencyStop = isEStop;
+                    decoder->direction = dir;
+
+                    decoder->throttle = throttle;
+
+                    for(int i = 0; i <= maxFunc; i++)
+                    {
+                      decoder->setFunctionValue(i, val[i]);
+                    }
                   }
+                }
+                catch(...)
+                {
+
                 }
               });
           }
--- a/server/src/hardware/protocol/z21/clientkernel.cpp
+++ b/server/src/hardware/protocol/z21/clientkernel.cpp
@@ -148,10 +148,15 @@ void ClientKernel::receive(const Message& message)
               targetSpeedStep = float(targetSpeedStep) / float(cache->speedSteps) * 126.0;
             }
 
-            if(cache->lastReceivedSpeedStep <= currentSpeedStep)
-              cache->speedTrend = LocoCache::Trend::Ascending;
-            else
-              cache->speedTrend = LocoCache::Trend::Descending;
+            if(!cache->speedTrendExplicitlySet)
+            {
+              //Calculate new speed trend
+              if(cache->lastReceivedSpeedStep <= currentSpeedStep)
+                cache->speedTrend = LocoCache::Trend::Ascending;
+              else
+                cache->speedTrend = LocoCache::Trend::Descending;
+            }
+            cache->speedTrendExplicitlySet = false;
 
             if(reply.speedSteps() != cache->speedSteps || reply.speedStep() != cache->speedStep)
             {
@@ -592,6 +597,7 @@ void ClientKernel::decoderChanged(const Decoder& decoder, DecoderChangeFlags cha
           if(cache->lastReceivedSpeedStep < newTargetSpeedStep)
             cache->lastReceivedSpeedStep = 126; //Reset to maximum
         }
+        cache->speedTrendExplicitlySet = true;
 
         //Update last seen time to ignore feedback messages of our own changes
         //This potentially ignores also user commands coming from Z21 if issued
--- a/server/src/hardware/protocol/z21/clientkernel.cpp
+++ b/server/src/hardware/protocol/z21/clientkernel.cpp
@@ -333,49 +333,21 @@ void ClientKernel::decoderChanged(const Decoder& decoder, DecoderChangeFlags cha
   {
     LanXSetLocoDrive cmd;
     cmd.setAddress(decoder.address, decoder.protocol == DecoderProtocol::DCCLong);
+    cmd.setDirection(decoder.direction);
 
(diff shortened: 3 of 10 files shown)


gfgit:
Should have kept it in Draft state...
`LanXLocoInfo` checksum code needs a bit of cleaning but I don't know whats best to do.

reinder:
Wouldn't it be better to cache these in the Z21 Kernel:
```
Direction lastCSDirection = Direction::Forward;
float lastCSThrottle = 0;
```
Then it can be tested in the kernel thread, if the is no change, there is no need to notify the eventloop. It can also be used the other way, if the eventloop send an instruction but it is already set, then it can be dropped.

For `LanXLocoInfo` may suggestion is to add a `isChecksumValid` method to `LanXLocoInfo` and create a new function for checksum validation e.g. `isChecksumValid(const LanX& message)` that function can call `LanXLocoInfo.isCheckSumValid` or the standard `XpressNet::isChecksumValid`. The new function should then be used for all LanX checksum validation, we then have a single place to handle this difference :) (and easily can other in the future if necessary)

gfgit:
> It can also be used the other way, if the eventloop send an instruction but it is already set, then it can be dropped.

Could you explain this better?
Also wouldn't this cache be desiderable also for other interfaces?
I would like a more centralized placet for that. Like `DecoderController` class with it's own logic maybe.
As for the eventloop yes it would be better to check in kernel thread.

gfgit:
> LocoNet does this

I cannot find this piece of code you mention

reinder:
here: https://github.com/traintastic/traintastic/blob/master/server/src/hardware/protocol/loconet/kernel.cpp#L706

I now discover that I broke it when moving most function update stuff to `updateFunctions` there the if changed is missing.

gfgit:
> here: https://github.com/traintastic/traintastic/blob/master/server/src/hardware/protocol/loconet/kernel.cpp#L706

And this is not called by `decoderChanged` because relies on command stations echoing back the message right?
So slot and decoder keep synced

gfgit:
Also I discovered Z21 message checksum are not validated at all except for LAN_X family

reinder:
> And this is not called by `decoderChanged` because relies on command stations echoing back the message right?
> So slot and decoder keep synced

Yes that is correct, on LocoNet there is always an "echo" because it receives the message it sends. However it would be wise I thing to read the slot every now and the just to make sure they are in sync. I'd like to do some traffic optimization in the future, LocoNet isn't very fast and we don't want to flood the bus. At my local club they are building a big layout that is operated using LocoNet, that will be a good test case in the future for Traintastic :)

reinder:
> Also I discovered Z21 message checksum are not validated at all except for LAN_X family

Only the `LAN_X` messages have a checksum, because it is a XpressNet message wrapped in a Z21 message and the XpressNet message has a checksum.

gfgit:
> At my local club they are building a big layout that is operated using LocoNet, that will be a good test case in the future for Traintastic :)

Wow! Hope to see it one day.

gfgit:
I'm getting new async loops (word of the year!) with Trains while Z21 is connected.
I suspect this is because traintastic sets loco speed and Z21 echos back.
I cannot detect the echo and ignore it because you cannot distinguish from a genuine handset controller message setting loco to that precise speed...

gfgit:
Also trains wit 5+ loco have problems because only first 3 loco get update and the issue seems on traintastic side judging from Wireshark capture

gfgit:
And there is a fight at startup to set track power state.
```
1. Traintast power off -> Z21
2. Z21 power on -> Traintastic power on -> Send Z21 skipped (same state)
3. Z21 in response to 1 sends "BC voltage off" -> Traintastic powers off -> Send Z21 skipped
```

But I think it goes on for a bit more. Other CS could suffer from this quick on/off

gfgit:
Maybe we could wait a bit before sending our state. Read received state from Z21, compare with ours then keep the more severe es "shortcircuit" should win over "run" and apply to both Z21 and World.
This way we prevent unwanted power on when shortcircuit happened and in general prevent value changing frequently.
Maybe we can add an andvanced option to turn off this feature

reinder:
> I'm getting new async loops (word of the year!) with Trains while Z21 is connected.
> I suspect this is because traintastic sets loco speed and Z21 echos back.
> I cannot detect the echo and ignore it because you cannot distinguish from a genuine handset controller message setting loco to that precise speed...

As long as the message speed is identical to the one set we should ignore it I think. Traintastic and the CS are on the same page then.

If a handset sets the same speed that is also fine. The they are all three on the same page.

gfgit:
> If a handset sets the same speed that is also fine. The they are all three on the same page.

What if in between the last speed set and the new speed received (possibly from handset or other source) traintastic has set a new speed?
I'm thinking about train Acceleration logic which updates speed many times in a few seconds if not less than a second. This could happen right in the middle of 2 different Z21 messages

gfgit:
Otherwise we could add a timestamp. If the received speed message happen less than 100ms after last speed set, we assume it's Z21 feedback and ignore it. If it was coming from an handset, user will probably move the throttle again and new message would be generate out of the timeout so it will be processed normally.
This is not very elegant solution because timeout is an arbitrary value so let's think something more clever

gfgit:
I've noticed Z21 feedback is received even after other speed was set, like:
- Set speed 1
- Set speed 2
- Feedback 1
- Feedback 2

So comparing received value with some "last set speed" to detect and ignore feedback might not work because it compares
second speed value with first feedback

gfgit:
> For `LanXLocoInfo` may suggestion is to add a `isChecksumValid` method to `LanXLocoInfo` and create a new function for checksum validation e.g. `isChecksumValid(const LanX& message)` that function can call `LanXLocoInfo.isCheckSumValid` or the standard `XpressNet::isChecksumValid`. The new function should then be used for all LanX checksum validation, we then have a single place to handle this difference :) (and easily can other in the future if necessary)

Done!

Now it works also with trains but there is still a problem if you explicitly set decoder steps to 128.
This is because `DecoderThrottleProperty` will round it to multiples of 1/128 while Z21 uses 126 steps so you get rounding errors.
This triggers train logic into thinking speed has changed while in reality it didn't, it's just different rounding.
I'm considering removing `DecoderThrottleProperty` and avoid rounding completely, like you suggested.
This will make try/catch blocks unneeded but may still be a good thing.

I've partially fixed the problem by saving last Z21 speed step and compare it.
I know `const_cast<>` is really ugly, it will be reworked after testing proves this patch is ready

reinder:
I think we should extend the `DecoderController` a bit by adding a list of valid speedsteps and protocols. The we can easily limit the values to choose from based on the protocol. A bit similar to the channel feature of the `InputController` and `OutputController`. For systems supporting only one protocol/speedsteps we can hide those controls. 

I'm still a bit in doubt about the float throttle vs. speed step control. Using speedsteps makes it possible to do better speed profiling and the acceleration/deceleration system can be a bit improved, instead of updating every 100ms, it can calculate the time between the steps, that should smooth it a bit more. This will however move the speedstep vs float issue to the vehicle as it needs to do the km/h to step conversion.

@gfgit what do you think?

> I'm considering removing DecoderThrottleProperty and avoid rounding completely, like you suggested.

I think we should keep it a float (without rounding) or just use speed step.

> I've partially fixed the problem by saving last Z21 speed step and compare it.

In the LocoNet implementation I only update the throttle value if the it results in a different speed step, see: [updateDecoderSpeed() in loconet/kernel.cpp](https://github.com/traintastic/traintastic/blob/master/server/src/hardware/protocol/loconet/kernel.cpp#L47)

gfgit:
> I think we should extend the `DecoderController` a bit by adding a list of valid speedsteps and protocols. The we can easily limit the values to choose from based on the protocol. A bit similar to the channel feature of the `InputController` and `OutputController`. For systems supporting only one protocol/speedsteps we can hide those controls.

This sound good!

> I'm still a bit in doubt about the float throttle vs. speed step control. Using speedsteps makes it possible to do better speed profiling and the acceleration/deceleration system can be a bit improved, instead of updating every 100ms, it can calculate the time between the steps, that should smooth it a bit more. This will however move the speedstep vs float issue to the vehicle as it needs to do the km/h to step conversion.

In my opinion the conversion is already done by rail vehicle, it's just we haven't yet implemented speed curves.
If you change to storing steps, you still cannot use them for acceleration because not all vehicles in a train must have same speed step range. You can couple an old 28 steps decoder to a newer 128 steps, the train cannot work in either range so it must work on speed value and then each rail vehicle will convert it to appropriate step.
The only thing here is if we go `float speed (km/h) -> float throttle (0-1) -> int step (0-n)` or directly `float speed (km/h) -> int step (0-n)`

gfgit:
Also do you think when getting decoder feedback from Z21 it should allow falling back to `Auto` protocol?
On decoder change it doesn't check protocol: we send everything as it were DCC.
I think Z21 also supports Motorola but don't have anyway to test it.
From what I understand motorola support is less complete than DCC, like you do not get feedback from Z21 for motorola decoders.

gfgit:
To workaround async Z21 feedback being processed after 2 send message, I've tried storing **last two** sent speed step and compare either one with received step.
This is really ugly code but it's the only thing that I can make work.
I Z21 is slow or there is network delay, I don't know if this holds.

gfgit:
Just by adding 500ms delay to each message on Z21 side it triggers infinite loop for everything Train-related.
Our logic to ignore Z21 feedback must not depend on time

gfgit:
Rebased on top of #64

gfgit:
New idea: we add add a timeout of 2 seconds in which we ignore any loco commands after every command sent.

- Add a `LocoCache` caching to Z21 `ClientKernel` similar to LocoNet Slot system
- Store time of last command issued by traintastic (So inside `decoderChanged()`)
- When we receive `LAN_X_LOCO_INFO` compare with store time, if less than 2 seconds passed ignore the message
  Chances are these messages are feedback for our `LAN_X_LOCO_SET_DRIVE`
- If message is an emergency stop do not ignore it.
- Ideally the timeout should be applied only when vehicle is in active train and train is accelerating/decelerating
  These 2 situation generate many messages in short period of time and trigger the "async loop".
- Do not store time of last Z21 command as we do not want to ignore user command even if sent shortly after one another

This does not yet work, I'm having an hard time investigating it inside WireShark

gfgit:
It works by increasing timeout to 6 seconds.
Maybe we can do an internal test, like comparing time between 2 received messages and calculate a good value for timeout.
Timeout unfortunately is also applied to user commands if sent too quickly, which is very bad.
This happens because `decoderChanged()` is called after a bunch of user changes.
This means when it's called it compares it's status (old) with the `LocoCache` status (new) even though it was triggered by an older `LocoCache` change.
This does not loop as before but still sets the timeout so a new user command gets ignored.
To better understand:
```
1. LocoCache speed 0, Decoder speed 0
2. Z21 (User form manual throttle) set speed 1
3. LocoCache speed 1 -> queues 'EventLoop::call()' to update decoder (still at 0)
4. Z21 (User form manual throttle) set speed 2
5. LocoCache speed 2 -> queues 'EventLoop::call()' to update decoder (still at 0)
6. EventLoop executes our callback -> Decoder speed = 1 -> It changed -> post to kernel (already at 2)
7. LocoCache receive speed 1 from EventLoop -> different -> send to Z21 speed 1 and set timeout
8. Z21 (feedback from out LOCO_SET_DRIVE) set speed to 1 -> ignore because within timeout
9. EventLoop executes our second callback -> Decoder speed = 2 -> It changed -> post to kernel (now at 1 again)
10. LocoCache receive speed 2 from EventLoop -> different -> send to Z21 speed 2 and set timeout again
11. Now everyone agrees on speed = 2 but we set timeout so new user command will be ignored for the next 6 seconds
```

In short:

| Operation | Z21 | LocoCache | Decoder |
|-----------|-----------|---------|-----|
| 1             |      0        |     0      |  0    |
| 2             |      1        |    0       |  0    |
| 3             |      1        |    1       |  0    |
| 4             |      2        |    1       |  0    |
| 5             |      2        |    2       |  0    |
| 6             |      2        |    2       |  1    |
| 7             |      2        |    1       |  1    |
| 8             |      1        |    1       |  1    |
| 9             |      1        |    1       |  2    |
| 10           |      1        |    2       |  2    |
| 11           |      2        |    2       |  2    |

Note the oscillation of Z21 speed which is not desiderable

gfgit:
By just not propagating the changes back to Z21 when coming from itself we achieve a good result:

PROS
- Train acceleration does not seem to trigger loops
- User can set a new speed while train is accelerating/decelerating, there is a quick flickering (throttle value oscillates a bit)
  But soon train gets the new target speed and acceleration continues smoothly.
- E-Stop and direction changes are always processed correctly
- Setting speed to zero from external throttle (not E-Stop) works, it triggers smooth deceleration

CONS
- Small flickering (throttle value oscillation) when changing speed from external throttle while train is accelerating/decelerating.
  Other decoders in the train do not suffer from this flickering so the problem is small.
  Might still be evident when manually controlled loco has a sound decoder which might brake and then accelerate quickly.
- Setting speed to zero (not E-Stop) form Traintastic Client lowers starts deceleration but the halts soon after.
  So clicking "0" results in lowering speed down just a bit. This seems caused by a timing issue which does not seem to be present when setting to 0 from external throttle.

By also adding timeout:
PROS
- Setting speed to 0 from Client works again

CONS
- During delay decoder speed is out of sync with Z21 so bad things may happen
- During delay user cannot change speed from external throttle and see train adjusting accordingly
- When moving a train by slowly rotating external throttle (which sends multiple speed messages increasing gradually), only first one is considered and then timeout triggers and forces the rest to be ignored -> Out of sync

While first solutions seems better, I suspect it's just a lucky case. Behavior of Client setting to 0 and external throttle setting to 0 should be the same but there is a timing issue, this suggest that under heavy load / slow network other timing issues may arise and behavior would not be the same

gfgit:
In fact simulating a small delay on Z21 side (100ms for every message processed) is sufficient to trigger loops.
Enabling also timeout (at least 1 second) mitigates a bit by oscillating for a while but loop is not infinite anymore.
Timeout of less than 1 second still allows for infinite loops to trigger.

reinder:
I just read everything, nice analysis :) Hopefully I've some time later tonight to reply.

reinder:
> By just not propagating the changes back to Z21 when coming from itself we achieve a good result:

I think that is best for most cases, in the LocoNet kernel, `setValueInternal` is used to update the `throttle`, when using `setValueInternal` the Propery's `onChanged` (assigned at construct) isn't fired, so the `decoderChanged` won't be called too. The property changed signal is always fired.

When there are many changes "in the air" that is always a problem.

Another thing that might help is when a speed change is send, wait for the response before sending another one. (it is UDP so your might not get an answer or it might not be delivered)

I still think we should drop the `throttle` property and add available protocols/speedsteps to the `DecoderController`, it basically moves the issue to the RailVehicle but it makes it easier to implement speed tables and the `RailVehicle` then known about a bit more about the underlaying hardware limitations. (Like I meantioned earlier, do still like that?)

reinder:
@gfgit I created #72 for the `DecoderController` adjustments, work is on a separate branch. (I should do for all new features until finished)

gfgit:
> I still think we should drop the `throttle` property and add available protocols/speedsteps to the `DecoderController`, it basically moves the issue to the RailVehicle but it makes it easier to implement speed tables and the `RailVehicle` then known about a bit more about the underlaying hardware limitations. (Like I meantioned earlier, do still like that?)

I don't know about it. I think we should wait for at least 2 control station to work properly and then think about speed table.

> Another thing that might help is when a speed change is send, wait for the response before sending another one. (it is UDP so your might not get an answer or it might not be delivered)

We can keep track of sent messages and "tick" them when corresponding reply is received, otherwise we can retry sending or give up after a timeout. See #57

gfgit:
Rebased on current master.

@reinder Logger gets called also in Z21 kernel thread in some places, is it correct? Or should we get rid of `Eventloop::call()` which then calls Log inside `receive()` and `send()` and instead call Log directly?

gfgit:
Replaced by #82

PR #64, opened 2023-06-10
Some include fixes and forward declaration

gfgit:


reinder:
Regarding the `method.hpp` split, I think it is best to keep the copyright statement the same in the `tpp` as in the `hpp`, the code doesn't actually change it is just moved from one file to another.

gfgit:
> Regarding the `method.hpp` split, I think it is best to keep the copyright statement the same in the `tpp` as in the `hpp`, the code doesn't actually change it is just moved from one file to another.

Done

reinder:
Seems there are includes missing for the Release server build, can you add those, then we can merge it :)

gfgit:
Yes, the problem is that on My pc everithing builds fine, even tests with MinGW 64 so I don't know why there is a different behaviour.
Maybe it's because I use a mix of MinGW compile and LLD linker but also Clang build seems to fail

reinder:
I built the `Release` config on Ubuntu 22.04, added a few includes of `method.tpp` solves the issue.

Git patch file:  [include-fixes.patch.txt](https://github.com/traintastic/traintastic/files/11769450/include-fixes.patch.txt) (uploading a .patch file isn't allowed)

reinder:
Seems that `method.tpp` must be added to `trayicon.cpp`, hope that is the last one.

gfgit:
Done and rebased on current master

Issue #60, opened 2023-05-11
Z21 work around packet loss


gfgit:
Z21 uses UDP protocol, on top of this some implementations (mainly Arduino based one) could be not enough performant to keep up
the message stream and some packets can be lost.
Now imagine triggering STOP because you see some obstacle on the tracks but the packet gets lost...

I suggest we implement some sort of queue of replies waiting to be received.
Each message type will have a sort of priority, higher priority means shorter timeout before sending again and higher retry count.
For example:
```
- Send LAN_X_SET_STOP (high priority)
- Wait 0.5 seconds for `LAN_X_BC_STOPPED`
- If not received retry 5 times.
```
Or:
```
- Send LAN_X_SET_LOCO_FUNCTION (low priority)
- Wait 1 second for `LAN_X_LOCO_INFO`
- If not received retry 2 times.
```

While waiting, other message can be sent/received because protocol is explicitly asynchronous.
Some received messages could look like our reply but are instead generated by other handset controllers connected to Z21.
So we need some logic to track messages, remove them from reply queue when we get a match and schedule retry when timeout expires.

Further optimization would be to use a single timer for all this logic and rework keep alive timer to only trigger when there has not been traffic since n seconds (configurable) but to not trigger when there is already a lot of traffic.

gfgit:
We could show some sort of error message to the user when Z21 seems not to receive our messages

reinder:
Yeah I like this, good suggestion! With UDP packet loss/drop is something to take into account indeed.

> While waiting, other message can be sent/received because protocol is explicitly asynchronous.

I think we should limit the number of messages "in flight" (advanced setting), in case of packet drops we don't want to flood it so more packets are dropped.

> Further optimization would be to use a single timer for all this logic and rework keep alive timer to only trigger when there has not been traffic since n seconds (configurable) but to not trigger when there is already a lot of traffic.

That would be nice, the Traintastic DIY does that too.

p.s. I've contacted the Z21 manufacturer to ask if the will support TCP (in the future), the sadly don't :( for computer control TCP would be much better. (we could still add it, maybe someday someone will implement it)

p.s.2. *advanced settings* in the future I think it would be good to label some settings as advanced, they can be hidden by default in the client, to make it easier for new users.

gfgit:
Well I've discovered Z21 protocol specification allows more messages to be packed in a single UDP datagram.
[...]

PR #73, opened 2023-06-24
Z21 pending request queue

gfgit:
Pending request queue allows to retry sending messages after a timeout if expected reply is not received.
This also allows to detect a message is a reply to our own changes and react to it differently

See #60 for earlier discussion.

NOTE: needs #82 applied first

Both this and #82 `LocoCache` try to resolve the "async loop" issue (See #57).
This 2 solutions are orthogonal and can live together.

### DIFF
--- a/server/src/hardware/protocol/z21/messages.cpp
+++ b/server/src/hardware/protocol/z21/messages.cpp
@@ -375,6 +375,7 @@ MessageReplyType getReplyType(const Message &message)
       reply.setPriority(MessageReplyType::Priority::Low);
       return reply;
     }
+
     case LAN_GET_BROADCASTFLAGS:
     case LAN_LOCONET_DETECTOR:
     case LAN_CAN_DETECTOR:
--- a/server/src/hardware/protocol/z21/messages.cpp
+++ b/server/src/hardware/protocol/z21/messages.cpp
@@ -496,7 +496,8 @@ MessageReplyType getReplyType(const Message &message)
                 setLocoDrive.db0 >= 0x10 && setLocoDrive.db0 <= 0x13)
             {
               reply.address = setLocoDrive.address();
-              reply.speedStep = setLocoDrive.speedStep();
+              reply.speedAndDirection = setLocoDrive.speedAndDirection;
+              reply.setSpeedSteps(setLocoDrive.speedSteps());
               reply.setFlag(MessageReplyType::Flags::CheckSpeedStep);
             }
             else if(const auto& setLocoFunction = static_cast<const LanXSetLocoFunction&>(message);
--- a/server/src/hardware/protocol/z21/messages.cpp
+++ b/server/src/hardware/protocol/z21/messages.cpp
@@ -361,4 +361,189 @@ bool LanX::isChecksumValid(const LanX &lanX)
   return XpressNet::isChecksumValid(msg, dataSize);
 }
 
+MessageReplyType getReplyType(const Message &message)
+{
+  switch (message.header())
+  {
+    //Messages whose replies have same header
+    case LAN_GET_SERIAL_NUMBER:
+    case LAN_GET_CODE:
+    case LAN_GET_HWINFO:
+    {
+      MessageReplyType reply;
+      reply.header = message.header();
+      reply.setPriority(MessageReplyType::Priority::Low);
+      return reply;
+    }
+    case LAN_GET_BROADCASTFLAGS:
+    case LAN_LOCONET_DETECTOR:
+    case LAN_CAN_DETECTOR:
+      return {message.header()};
+
+    case LAN_GET_LOCO_MODE:
+    case LAN_GET_TURNOUTMODE:
+    {
+      MessageReplyType reply;
+      reply.header = message.header();
+      reply.setFlag(MessageReplyType::Flags::CheckAddress);
+
+      if(message.header() == LAN_GET_LOCO_MODE)
+        reply.address = static_cast<const LanGetLocoMode&>(message).address();
+
+      // NOTE: not (yet) supported
+      //else
+      //  reply.address = static_cast<const LanGetTurnoutMode&>(message).address();
+
+      return reply;
+    }
+
+    // NOTE: This message has no reply for Z21 firmware < 1.22
+    // NOTE: not (yet) supported
+    case LAN_LOCONET_DISPATCH_ADDR:
+      return {LAN_LOCONET_DISPATCH_ADDR};
+
+    case LAN_X:
+    {
+      const LanX& lanX = static_cast<const LanX&>(message);
+      switch (lanX.xheader)
+      {
+        case 0x21:
+        {
+          MessageReplyType reply;
+          reply.header = LAN_X;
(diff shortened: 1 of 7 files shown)


gfgit:
Quick comparison with #59 approach (not fully accurate):

## **LocoCache** approach
PROS:

- Less memory footprint when number of loco is low and many messages are sent
- Can extract individual changes from `LanXLocoInfo` like change only direction or only emergency stop
  This is possible because we cache last value of direction, speed and emergency stop per each loco.
- Can somewhat detect speed trend change: user reduces speed while train is accelerating or viceversa

CONS:

- Weak logic based on heuristic. Many edge cases
- Logic is more obscure
- More memory when many loco are running
- Annoying "ignore" delay when user moves throttle right after Traintastic has updated same loco
  This is needed to avoid receiving our own reply but creates discrepancy between real loco state and cached one.

## Pending Request Queue approach

PROS:
- While logic is more complex, it's easier to understand
-  More accurate for own replies (which get then ignored)
- Expandable to other message types
- Allows re-sending a message if no reply is received so it workarounds UDP packet loss
- Logic is much more robust
- No delay needed for throttle messages

CONS:
- Can detect external changes as our own replies if messages match.
  This shouldn't be a big problem but there are at least some edge cases (see below)
- Potential risk of big memory footprint is sending a lot of messages in a fast manner
- Does not check received message order so 2 replies referring to same object can be
  detected in wrong order.
- Cannot extract individual changes because it has no track of last value

## Future work

The solution of course will be "best of both worlds". I'll try to tune both approaches to work together
and understand which one is doing what exactly and what are the edge cases left.

It already shows big improvement when simulating enormous network delays (100 ms for every message on Z21 side)
but it's not yet perfect.
It really needs testing with real hardware (which unfortunately I don't have) because different behaviors might arise

reinder:
Looking good, I really like it. Nice work! I don't own a z21 or Z21 yet so can't test it against a real one, I only have a Digikeijs DR5000 which supports the Z21 protocol.

> If you get speed 15 after speed 5 it means someone else has explicitly set speed to 15

Or that 6 .. 14 are dropped, it's UDP after all. But under normal circumstances without a high network load it probably won't happen.

Maybe you can add a debug option to log some additional info about the retries, that might be useful if we find people to do some testing with it.

p.s. As the Z21 implementation is receiving many improvements, it seems fair to me that you add a Copyright line for your name. (Don't know how this normally works...this is my first Open Source project with contributions :))

gfgit:
Rebased onto #82

server/src/hardware/protocol/z21/messages.hpp
```diff
@@ -1423,6 +1444,69 @@ constexpr std::string_view toString(LanXSetLocoFunction::SwitchType value)
...
 }
 
+struct MessageReplyType
+{
+  enum Priority
+  {
+    LowPriority = 0,
+    Normal = 1,
+    Urgent = 2
+  };
```
reinder, line 1454: Why does one element have `Priority` in it and the other don't have it?

Change it to `enum class Priority : uint8_t`? Then the values are scoped :)

server/src/hardware/protocol/z21/messages.hpp
```diff
@@ -1423,6 +1444,69 @@ constexpr std::string_view toString(LanXSetLocoFunction::SwitchType value)
...
+
+  enum Flags
```
reinder, line 1456: `enum class`?

server/src/hardware/protocol/z21/messages.hpp
```diff
@@ -1423,6 +1444,69 @@ constexpr std::string_view toString(LanXSetLocoFunction::SwitchType value)
...
+  {
+    CheckDb0       = 1,
+    CheckSpeedStep = 1 << 1,
+    CheckAddress   = 1 << 2
+  };
+
+  inline Priority getPriority() const
```
reinder, line 1463: remove `get` for getter, the Traintastic codebase style doesn't do that.

server/src/hardware/protocol/z21/messages.hpp
```diff
@@ -1423,6 +1444,69 @@ constexpr std::string_view toString(LanXSetLocoFunction::SwitchType value)
...
+  {
+    return Priority(m_flags & 0xF);
+  }
+
+  inline void setPriority(Priority priority)
```
reinder, line 1468: as the above is change, the argument name causes a *shadow* warning, that's why the common practice in the Traintastic codebase is to use `value` as name for the setter argument.

reinder (changes requested):

server/src/hardware/protocol/z21/messages.cpp
```diff
@@ -324,4 +324,187 @@ bool LanX::isChecksumValid(const LanX &lanX)
   return XpressNet::isChecksumValid(msg, dataSize);
 }
 
+MessageReplyType getReplyType(const Message &message)
+{
+  switch (message.header())
```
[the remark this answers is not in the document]
reinder: yeah, clang-format would really help, I made one some time ago, that was working quiet well. There were a few issues, please remember me again if I forget :)

gfgit:
I now have built a Z21 with ESP32 and did some initial tests. Pending reply queue is the way to go.
We will keep some bits of my old `LocoCache` and remove it's timed logic and speed trend logic which never worked right and are difficult to understand.
Also `LocoCache` should get a better name.
I will clean the code and do some more tests

gfgit:
Last commit is a fix unrelated to this PR but found when testing.
Also currently you cannot really test this PR in simulation mode because IO handler will not reply to locomotive drive messages, this is detected as UDP packet drop so messages are sent again 2 times.

gfgit:
@reinder What do you think about this?

reinder:
> Also currently you cannot really test this PR in simulation mode because IO handler will not reply to locomotive drive messages, this is detected as UDP packet drop so messages are sent again 2 times.

Can we add that in the simulation?

gfgit:
> > Also currently you cannot really test this PR in simulation mode because IO handler will not reply to locomotive drive messages, this is detected as UDP packet drop so messages are sent again 2 times.
> 
> Can we add that in the simulation?

Yes we could but since we do not cache decoder state in simulation IO handler, the reply would contain incorrect values.

gfgit:
The problem is that reply is always a `LAN_X_LOCO_INFO` which contains all loco state. So if we rely to decoder throttle change we need to put in also functions state which we do not know because we do not cache.
By putting them all to off we would break behavior of throttle widget for example because function would stay on only until a new reply is received.
If you want I can make something similar to LocoChache inside simulation IO handler to store decoder state for replies

gfgit:
Proof of concept. Not tested yet

reinder:
Looking good, it is really helpful to have proper simulation, this makes it easier to do development without hardware attached :) (We still need to test  with real hardware off course)

reinder (approved):

reinder:
@gfgit Just tested simulation mode, works nicely. Anything else we need to check? (Or can I merge it?)

gfgit:
You can merge it!

PR #76, opened 2023-07-07
Upgrade to C++20

reinder:


reinder:
Not a good approach, so do it again in another branch and the just change the C++ standard, `tcb::span` stuff can be removed later, make it much easier to keep up with `master`.

Issue #53, opened 2023-04-13
Server hang on communication disabling


gfgit:
**Describe the bug**
Simple world with Z21 connected

**To Reproduce**
1. Open attached world
2. While disconnected from Z21, power on and run
3. Put Z21 in Emergency Stop state
4. Open Server Log view
5. Enable communication
6. You will see world going in stop and possibly a Broadcast flag mismatch warning
7. Disable communication
8. Server hangs for a socket error, something like "connection forcibly interrupted"?

**Expected behavior**
No hang

**World export**
[Test Local Z21 2023-04-13_new.zip](https://github.com/traintastic/traintastic/files/11224840/Test.Local.Z21.2023-04-13_new.zip)

**Traintastic server/client:**
 - OS: Windows 10
 - Version: last commit of PR #51
 - Compiler: MinGW

**Connected hardware**
Z21ServerEmulator running on localhost

**Additional context**
A socket exception is catched on mani() so no backtrace. And it is not happening always...

gfgit:
Found the issue but I need suggestions.
Z21's `ClientKernel` internally calls: (NOTE: lambda capture `this` to access `this->m_logId`)
```
EventLoop::call(
        [this, msg=toString(message)]()
        {
          Log::log(m_logId, LogMessage::D2001_TX_X, msg);
        });
```

When communication are disabled, kernel shared pointer is reset -> `ClientKernel` object deleted.
But event loop sometimes still has to process the lambda above so `m_logId` is accessed after kernel is deleted.

A possible solution is to get a shared pointer inside the lambda to avoid it being deleted while lambda is executed.
This should be done for every lambda and also for other command stations which might present same race behaviour.

Drawback is that Kernel object might have connections inside it's parent hardware interface and the interface expect to have deleted the kernel immediately but in reality it might still use some callbacks or access some interface members

gfgit:
In this particolar case we could just capture a copy of log ID and not capture `this` at all.
But in other lambdas a reference to `this` is needed and we should use a different method.
Also I'm not sure `shared_ptr` is thread safe

gfgit:
This problem is partially fixed inside #59 but other lambdas passed to `EventLoop::call(...)` are capturing `this` so we need to change them also.
Other interfaces might have the same issue.

reinder:
[...]

PR #81, opened 2023-11-02
New Z21 fixes

gfgit:
Hi! These changes are extracted from other PRs, mainly #59 and #73, #56 and do not change behavior except bug fixing.

- d9ace24d88c32cd768a6e93ee4e17280558803b2 First commit is a short term solution to #53 but can also be kept afterwards as it's better to capture as less as possible.
- 0e463ff8fe97889b0921bda4b519e8f28058d2ce Can be explained a bit better, test if you run in race condition by connecting station in a different world state than Traintastic and see who imposes it's state on the other.
- b48e8f75069e79d5c3e021e3d52c795c89f028a4 Double check format is now correct
- 488663331dbbc002491a8d9c5a72799659f3f85e There are still other hardcoded values in `simulationiohandler.cpp` and `serverkernel.cpp` which can be labelled but didn't know what names to give them.

### DIFF
--- a/server/src/hardware/protocol/z21/clientkernel.cpp
+++ b/server/src/hardware/protocol/z21/clientkernel.cpp
@@ -333,49 +333,21 @@ void ClientKernel::decoderChanged(const Decoder& decoder, DecoderChangeFlags cha
   {
     LanXSetLocoDrive cmd;
     cmd.setAddress(decoder.address, decoder.protocol == DecoderProtocol::DCCLong);
+    cmd.setDirection(decoder.direction);
 
-    switch(decoder.speedSteps)
+    // Decoder max speed steps must be set for the message to be correctly
+    // distinguished from LAN_X_SET_LOCO_FUNCTION
+    cmd.setSpeedSteps(decoder.speedSteps);
+
+    if(decoder.emergencyStop)
     {
-      case 14:
-      {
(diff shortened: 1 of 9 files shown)


server/src/hardware/protocol/z21/clientkernel.cpp
```diff
@@ -64,56 +63,53 @@ void ClientKernel::receive(const Message& message)
...
           if(message == LanXBCStopped())
           {
-            if(m_emergencyStop != TriState::True)
-            {
-              m_emergencyStop = TriState::True;
-
-              if(m_onEmergencyStop)
-                EventLoop::call(
-                  [this]()
+            if(m_onEmergencyStop)
```
reinder, line 103: move `if(m_onEmergencyStop)` into the lambda just like above?

reinder (commented):
Looking good!

server/src/hardware/protocol/z21/clientkernel.cpp
gfgit: So the `m_emergencyStop` flag should be updated even if no lamda is set right?

gfgit:
Fixed lamda and rebased on current master.

server/src/hardware/protocol/z21/clientkernel.cpp
reinder: Yes Indeed, just like with `m_trackPowerOn`, the lambda should always be set, but for consistency it is best to handle both in the same manner.

PR #82, opened 2023-11-02
Z21 handle loco info 2

### LINKED ISSUE #57: Z21 Async nature

gfgit:
Replaces #59

This PR handles receiving loco state from Z21 and update relevant Decoders.

Needs #81 first

- Added support for receiving F29 - F31 (When Z21 firmware version is >= 1.42)
- Fix infinite loop explained in issue #57 (only mitigated)

### DIFF
--- a/server/src/hardware/protocol/z21/clientkernel.cpp
+++ b/server/src/hardware/protocol/z21/clientkernel.cpp
@@ -121,10 +121,11 @@ void ClientKernel::receive(const Message& message)
           {
             const auto& reply = static_cast<const LanXLocoInfo&>(message);
 
-            const int maxFunc = reply.supportsF29F31() ? 31 : 28;
-            bool val[31 + 1] = {};
+            //NOTE: there is also a function at index 0, hence +1
+            const int functionIndexMax = std::min(reply.functionIndexMax(), LanXLocoInfo::supportedFunctionIndexMax);
+            bool val[LanXLocoInfo::supportedFunctionIndexMax + 1] = {};
 
-            for(int i = 0; i <= maxFunc; i++)
+            for(int i = 0; i <= functionIndexMax; i++)
             {
               val[i] = reply.getFunction(i);
             }
@@ -231,7 +232,7 @@ void ClientKernel::receive(const Message& message)
             EventLoop::call(
               [this, address=reply.address(), isEStop=reply.isEmergencyStop(),
               speed = reply.speedStep(), speedMax=reply.speedSteps(),
-              dir = reply.direction(), val, maxFunc, changes]()
+              dir = reply.direction(), val, functionIndexMax, changes]()
               {
                 try
                 {
@@ -262,7 +263,7 @@ void ClientKernel::receive(const Message& message)
 
                     //Function get always updated because we do not store a copy in cache
                     //so there is no way to tell in advance if they changed
-                    for(int i = 0; i <= maxFunc; i++)
+                    for(int i = 0; i <= functionIndexMax; i++)
                     {
                       decoder->setFunctionValue(i, val[i]);
                     }
--- a/server/src/hardware/protocol/z21/clientkernel.cpp
+++ b/server/src/hardware/protocol/z21/clientkernel.cpp
@@ -130,7 +130,7 @@ void ClientKernel::receive(const Message& message)
               val[i] = reply.getFunction(i);
             }
 
-            LocoCache *cache = getLocoCache(reply.address());
+            LocoCache &cache = getLocoCache(reply.address());
 
             DecoderChangeFlags changes = DecoderChangeFlags(0);
 
@@ -139,27 +139,27 @@ void ClientKernel::receive(const Message& message)
             if(reply.speedSteps() != 126)
             {
               currentSpeedStep = float(currentSpeedStep) / float(reply.speedSteps()) * 126.0;
-              if(abs(currentSpeedStep - cache->lastReceivedSpeedStep) < 5)
-                currentSpeedStep = cache->lastReceivedSpeedStep; //Consider it a rounding error
+              if(abs(currentSpeedStep - cache.lastReceivedSpeedStep) < 5)
+                currentSpeedStep = cache.lastReceivedSpeedStep; //Consider it a rounding error
             }
 
-            int targetSpeedStep = cache->speedStep;
-            if(cache->speedSteps != 126)
+            int targetSpeedStep = cache.speedStep;
+            if(cache.speedSteps != 126)
             {
-              targetSpeedStep = float(targetSpeedStep) / float(cache->speedSteps) * 126.0;
+              targetSpeedStep = float(targetSpeedStep) / float(cache.speedSteps) * 126.0;
             }
 
-            if(!cache->speedTrendExplicitlySet)
+            if(!cache.speedTrendExplicitlySet)
             {
               //Calculate new speed trend
-              if(cache->lastReceivedSpeedStep <= currentSpeedStep)
-                cache->speedTrend = LocoCache::Trend::Ascending;
+              if(cache.lastReceivedSpeedStep <= currentSpeedStep)
+                cache.speedTrend = LocoCache::Trend::Ascending;
               else
-                cache->speedTrend = LocoCache::Trend::Descending;
+                cache.speedTrend = LocoCache::Trend::Descending;
             }
-            cache->speedTrendExplicitlySet = false;
+            cache.speedTrendExplicitlySet = false;
 
-            if(reply.speedSteps() != cache->speedSteps || reply.speedStep() != cache->speedStep)
+            if(reply.speedSteps() != cache.speedSteps || reply.speedStep() != cache.speedStep)
             {
               // Use a timeout of 1 second to prevent reacting to Z21 feedback messages
               // of our own changes. This would be problematic because in the meatime our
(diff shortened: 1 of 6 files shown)


server/src/hardware/protocol/z21/clientkernel.cpp
```diff
@@ -572,4 +802,47 @@ void ClientKernel::keepAliveTimerExpired(const boost::system::error_code& ec)
...
+    else
+    {
+      it++;
+    }
+  }
+
+  startInactiveDecoderPurgeTimer();
+}
+
+ClientKernel::LocoCache *ClientKernel::getLocoCache(uint16_t dccAddr)
```
reinder, line 835: This method always returns a valid pointer, so it should return a reference.

When pointers are involved you always need to think about `nullptr`, in this case a reference eliminates this :)

server/src/hardware/protocol/z21/clientkernel.hpp
```diff
@@ -81,13 +84,52 @@ class ClientKernel final : public Kernel
...
+     * Z21::ClientKernel::onStart().
+     *
+     * \sa EventLoop
+     */
     TriState m_trackPowerOn;
+
+    /*!
+     * \brief m_emergencyStop caches command station emergency stop state.
+     *
+     * NOTE: it must be accessed only from event loop thread or from
```
reinder, line 100: Doxygen also supports `\note`

server/src/hardware/protocol/z21/clientkernel.hpp
```diff
@@ -81,13 +84,52 @@ class ClientKernel final : public Kernel
...
+      uint8_t speedSteps = 0;
+      uint8_t lastReceivedSpeedStep = 0; //Always in 126 steps
+      Trend speedTrend = Trend::Ascending;
+      bool speedTrendExplicitlySet = false;
+      Direction direction = Direction::Unknown;
+      std::chrono::steady_clock::time_point lastSetTime;
+    };
+
+    std::unordered_map<uint16_t, LocoCache> m_locoCache;
+    bool isUpdatingDecoderFromKernel = false;
```
reinder, line 131: rename to `m_isUpdatingDecoderFromKernel` it's a member variable

server/src/hardware/protocol/z21/messages.hpp
```diff
@@ -1251,16 +1292,24 @@ struct LanXLocoInfo : LanX
...
-    checksum = xheader;
-    for(uint8_t* db = &addressHigh; db < &checksum; db++)
-      checksum ^= *db;
+    //Data length - 7 Z21 header bytes + 1 byte for db0
+    LanX::updateChecksum(dataLen() - 6);
   }
 } ATTRIBUTE_PACKED;
-static_assert(sizeof(LanXLocoInfo) == 14);
+static_assert(sizeof(LanXLocoInfo) >= LanXLocoInfo::minMessageSize &&
+              sizeof(LanXLocoInfo) <= LanXLocoInfo::maxMessageSize);
```
reinder, line 1312: size of this struct is now always 15, the `static_assert` is only checked at compile time

server/src/hardware/protocol/z21/messages.hpp
```diff
@@ -1094,14 +1123,18 @@ struct LanXLocoInfo : LanX
...
+  uint8_t speedAndDirection = 0; //db3
   uint8_t db4 = 0;
-  uint8_t f5f12 = 0;
-  uint8_t f13f20 = 0;
-  uint8_t f21f28 = 0;
+  uint8_t f5f12 = 0;    //db5
+  uint8_t f13f20 = 0;   //db6
+  uint8_t f21f28 = 0;   //db7
+  uint8_t f29f31 = 0;   //db8 (firmware >= 1.42)
   uint8_t checksum = 0;
```
reinder, line 1138: Shouldn't these be renamed to `db8` and `db9` ? (and add a `checksum() const` method.

When the message is sent from a Z21 with firmware < 1.42 the `f29f31` field contains the checksum and the `checksum` field is then probably undefined or results in a memory access error, this is prone to bugs.

reinder (changes requested):
Nice improvement!s! Really want to test is on a real Z21, I might be able to borrow a black z21 in the near future. :)

reinder:
Build has same mingw issue, but that will be resolved with a rebase on the current master, as that is fixed in #81.

server/src/hardware/protocol/z21/clientkernel.cpp
```diff
@@ -64,58 +65,218 @@ void ClientKernel::receive(const Message& message)
...
+                  //When accelerating or decelerating ignore all speeds between original speed and target speed.
+                  //These messages are probably feedback of our own changes arrived with some delay.
+                  //If we get values outside this range or changing trend we pass them to let Train adjust.
+                  maybeOldFeedback = true;
+                }
+
+                if(!maybeOldFeedback)
+                {
+                  changes = changes | DecoderChangeFlags::Throttle;
+                  changes = changes | DecoderChangeFlags::SpeedSteps;
```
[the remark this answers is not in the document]
gfgit: I've added the `DecoderChangeFlag::operator|=()` which internally does the same.

server/src/hardware/protocol/z21/clientkernel.hpp
gfgit: Mind blowing!

gfgit: But it was part of #81 so it will be added on new commit

server/src/hardware/protocol/z21/messages.hpp
gfgit: Isn't it cool to have an abstract compile time logic? So it does not have to be changed when we change the struct layout

server/src/hardware/protocol/z21/messages.hpp
gfgit: I don't think we need a checksum method because we can already use `LanX::isChecksumValid(const LanX &lanX)` which has a special case for this message.
The `checksum` now renamed `db9` is not accessed anywhere

server/src/hardware/protocol/z21/clientkernel.cpp
```diff
@@ -64,58 +65,218 @@ void ClientKernel::receive(const Message& message)
...
+            //Do not update last seen time to avoid ignoring genuine user commands
+            //Store last received speed step converted to 126 steps scale
+            cache->lastReceivedSpeedStep = currentSpeedStep;
+
+            EventLoop::call(
+              [this, address=reply.address(), isEStop=reply.isEmergencyStop(),
+              speed = reply.speedStep(), speedMax=reply.speedSteps(),
+              dir = reply.direction(), val, maxFunc, changes]()
+              {
+                try
```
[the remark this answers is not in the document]
gfgit: Assignment to Decoder `throttle` property might throw an exception which doesn't get handled by `EventLoop` and thus crashes the server. I used to get these crashes only when code was still draft but it should do no harm to keep it.
Also when code throws I think it's correct to always put it inside try/catch blocks

server/src/hardware/protocol/z21/clientkernel.cpp
gfgit: We should guard against out of memory if cache gets too big...

server/src/hardware/protocol/z21/clientkernel.hpp
```diff
@@ -81,13 +84,52 @@ class ClientKernel final : public Kernel
...
+      uint16_t dccAddress = 0;
+      bool isEStop = false;
+      uint8_t speedStep = 0;
+      uint8_t speedSteps = 0;
+      uint8_t lastReceivedSpeedStep = 0; //Always in 126 steps
+      Trend speedTrend = Trend::Ascending;
+      bool speedTrendExplicitlySet = false;
+      Direction direction = Direction::Unknown;
+      std::chrono::steady_clock::time_point lastSetTime;
+    };
```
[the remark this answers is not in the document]
gfgit: I don't remember the logic exactly but it's different. Explicitly set is about overwriting existing value or not when receiving new input while Unset would represent a valid value without meaning to calculate trend.
Explicitly set trend disables overwriting for next receive call which disables the flag, allowing future overwriting.
This was needed to avoid interfering with Traintastic throttle acceleration.

Consider this scenario:

1.  Decoder throttle is 50
2. User with external throttle moves it to 100 (Ascending trend). Being it connected directly to Z21, Traintastic has no way to intercept this value before it gets processed.
3. Traintastic gets the feedback from Z21 and updates `decoder->throttle` property
4. If decoder is in a Train, this will trigger Train target speed change and begins acceleration curve
5. Acceleration curve starts at current Train speed so **decoder is put back** at `throttle = 50`
6. This value might seem a descending trend while is just reverting recent throttle update to begin real acceleration
7. Z21 sees new throttle 50, this is unavoidable and might cause problems with sound decoder as they will receive in rapid succession values 50, **100**, 50, 51, 52, ..., 100. There is a spike at the beginning.

It may sound complicated but it's not. If you have doubts, just tell me and I'll explain it better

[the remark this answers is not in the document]
gfgit: I've made this trend logic to distinguish feedback of our own triggered changes from genuine user throttle changes.
This method is not perfect and will detect false positives leading to user throttle changes being silently ignored.
The explicitly set logic mitigates Train acceleration algorithm from messing all up

gfgit:
Since some changes conflict, I decided to not rebase current commits and instead add new ones with fixes. These are added on top of #73 unfortunately

server/src/hardware/protocol/z21/clientkernel.hpp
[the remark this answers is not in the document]
reinder: Ok, we keep it like it is :)

server/src/hardware/protocol/z21/clientkernel.cpp
```diff
@@ -112,6 +113,44 @@ void ClientKernel::receive(const Message& message)
...
+
+        case LAN_X_LOCO_INFO:
+        {
+          const int n = message.dataLen() - 7;
+          if(n >= 7 && n <= 14)
+          {
+            const auto& reply = static_cast<const LanXLocoInfo&>(message);
+
+            bool val[31 + 1] = {};
+            const int maxFunc = (n >= 8) ? 31 : 28;
```
reinder: suggestion: add a method to `LanXLocoInfo` to get the max function number, if the message is extended again in the future, this will still work :)
And a static constexpr with the message maximum for the `bool val[]` array above :)

server/src/hardware/protocol/z21/clientkernel.cpp
reinder: True, but that is rare, cache won't be bigger that approx. ~10000 loco's (DCC address space) worst case, which still isn't a huge amount of memory.

Traintastic currently doesn't handle OS out of memory, it would be nice to stop all trains and power off in that case, however OS out of memory error are rare these days.

reinder (approved):

server/src/hardware/protocol/z21/messages.hpp
gfgit: I don't know if we already reched max size, using every byte field in the struct. Maybe there is still reserved space for future expansion.

Issue #94, opened 2023-12-18
Add Swedish translation


reinder:
Tasks:
- [x] Translate application strings
- [x] Translate Windows installer strings (InnoSetup)

reinder:
see #98

PR #96, opened 2024-01-11
Translations SV-SE

princeofnaxos:
Swedish translations

### DIFF
--- a/shared/translations/en-us.json
+++ b/shared/translations/en-us.json
@@ -833,7 +833,7 @@
     },
     {
         "term": "decoder_function_type:on_off",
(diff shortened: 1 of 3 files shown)


reinder:
Awesome :sunglasses:, I'll merged it. See #94.

Are you willing todo the installer translation also? (It are a few strings), see [en-us.isl](https://github.com/traintastic/traintastic/blob/master/package/innosetup/en-us.isl) and add one line in  [traintastic.iss](https://github.com/traintastic/traintastic/blob/master/package/innosetup/traintastic.iss#L41).

PR #108, opened 2024-04-09
Update Italian translation

gfgit:
Hi, I've taken all new terms from English file and also fixed a typo in it.
I think commits could be squashed into only a couple of commits total to reduce git history.
I did many commits just to make sure I didn't miss anything.


reinder:
Many thanks for the update @gfgit !

PR #113, opened 2024-04-16
Recognize output state

gfgit:
This is still WIP

I would like to get output tiles shown on Board to react to external state changes.
For example switching a turnout from command station or handset should result in turnout rail tile showing the new position.

How to test:

1. Create 2 turnouts and assign them the same address
2. Switching one turnout should result in the other moving too

### DIFF
--- a/server/src/board/tile/rail/decouplerrailtile.cpp
+++ b/server/src/board/tile/rail/decouplerrailtile.cpp
@@ -67,6 +67,11 @@ DecouplerRailTile::DecouplerRailTile(World& world, std::string_view _id)
 
   Attributes::addObjectEditor(deactivate, false);
   m_interfaceItems.add(deactivate);
+
+  outputMap->onOutputStateMatchFound.connect([this](DecouplerState value)
+    {
+      setState(value, true);
+    });
 }
 
 void DecouplerRailTile::worldEvent(WorldState worldState, WorldEvent worldEvent)
@@ -78,11 +83,12 @@ void DecouplerRailTile::worldEvent(WorldState worldState, WorldEvent worldEvent)
   Attributes::setEnabled(name, editable);
 }
 
-void DecouplerRailTile::setState(DecouplerState value)
+void DecouplerRailTile::setState(DecouplerState value, bool skipAction)
 {
   if(state != value)
   {
-    (*outputMap)[value]->execute();
+    if(!skipAction)
+      (*outputMap)[value]->execute();
     state.setValueInternal(value);
   }
 }
--- a/server/src/hardware/output/map/outputmap.cpp
+++ b/server/src/hardware/output/map/outputmap.cpp
@@ -39,6 +39,12 @@
 #include "../../../utils/displayname.hpp"
 #include "../../../utils/inrange.hpp"
 
+//TODO: needed for connecting value changed signal. Can be avoided?
+#include "../singleoutput.hpp"
+#include "../pairoutput.hpp"
+#include "../aspectoutput.hpp"
+#include "../ecosstateoutput.hpp"
+
 OutputMap::OutputMap(Object& _parent, std::string_view parentPropertyName)
   : SubObject(_parent, parentPropertyName)
   , interface{this, "interface", nullptr, PropertyFlags::ReadWrite | PropertyFlags::Store | PropertyFlags::NoScript,
@@ -123,16 +129,17 @@ OutputMap::OutputMap(Object& _parent, std::string_view parentPropertyName)
         }
         else // no interface
         {
-          for(auto& output : m_outputs)
+          for(auto& it : m_outputs)
           {
-            if(output) /*[[likely]]*/
+            if(it.first) /*[[likely]]*/
             {
-              interface->releaseOutput(*output, parent());
+              releaseOutput(it);
             }
-            m_outputs.clear();
-            addresses.clearInternal();
-            addressesSizeChanged();
           }
+
+          m_outputs.clear();
+          addresses.clearInternal();
+          addressesSizeChanged();
         }
         return true;
       }}
@@ -142,9 +149,9 @@ OutputMap::OutputMap(Object& _parent, std::string_view parentPropertyName)
         // Release outputs for previous channel:
         while(!m_outputs.empty())
         {
-          if(m_outputs.back()) /*[[likely]]*/
+          if(m_outputs.back().first) /*[[likely]]*/
           {
-            interface->releaseOutput(*m_outputs.back(), parent());
+            releaseOutput(m_outputs.back());
           }
           m_outputs.pop_back();
         }
@@ -209,18 +216,18 @@ OutputMap::OutputMap(Object& _parent, std::string_view parentPropertyName)
           return false; // Duplicate addresses aren't allowed.
         }
 
-        auto output = getOutput(channel, value, *interface);
-        if(!output) /*[[unlikely]]*/
+        auto outputConnPair = getOutput(channel, value, *interface);
+        if(!outputConnPair.first) /*[[unlikely]]*/
         {
           return false; // Output doesn't exist.
         }
 
-        if(index < m_outputs.size() && m_outputs[index])
+        if(index < m_outputs.size() && m_outputs[index].first)
         {
-          releaseOutput(*m_outputs[index]);
+          releaseOutput(m_outputs[index]);
         }
 
-        m_outputs[index] = output;
+        m_outputs[index] = outputConnPair;
 
         return true;
       }}
@@ -237,7 +244,7 @@ OutputMap::OutputMap(Object& _parent, std::string_view parentPropertyName)
         {
           if(!m_outputs.empty())
           {
-            releaseOutput(*m_outputs.front());
+            releaseOutput(m_outputs.front());
             m_outputs.clear();
           }
           if(value != 0)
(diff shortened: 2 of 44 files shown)


reinder:
Awesome! This is something that was still missing.

Another method of testing is using the output keyboard, turnout's etc. should also respond to that.

There are also some catches:
- Signal aspect is determined automatically based on the blocks/turnouts behind it.  So what todo when a change form the command station arrives? I think it is best to reevaluate the signal aspect and set the output(s) again.
- Turnouts can be reserved for a path, what to do if the turnout position changes to another position than required for the path?
- Decouplers are manually controlled only  for now, so that should be fine :)
-

server/src/board/tile/rail/decouplerrailtile.cpp
```diff
@@ -67,6 +67,11 @@ DecouplerRailTile::DecouplerRailTile(World& world, std::string_view _id)
 
   Attributes::addObjectEditor(deactivate, false);
   m_interfaceItems.add(deactivate);
+
+  outputMap->onOutputStateMatchFound.connect([this](DecouplerState value)
+    {
+      setState(value, true);
```
reinder, line 73: You can just do `state.setValueInternal(value);` here, because it is an internal update.

gfgit:
Thanks!


> * Signal aspect is determined automatically based on the blocks/turnouts behind it.  So what todo when a change form the command station arrives? I think it is best to reevaluate the signal aspect and set the output(s) again.

I was thinking about this too. But we need to be careful to not trigger recursion.
For example on double slip switch with 2 motors, so two outputs this happens:

1. Turnout is in "right" position: First, First
2. User switches to "left" inside Board, commands are sent to command station
3. Received feedback from first motor. Now state is: Second, First
4. Received feedback from second motor. Now state is: Second, Second. Operation completed.

The catch here is that while point 4 correctly matches the originally requested position, point 3 which is in the middle will match "Straight A" position because operation is not complete yet.

A similar thing could happen to a signal if represented with multiple outputs which do not switch together atomically.
So the partial match could trigger a reevaluation which would restart from point 1 

> * Turnouts can be reserved for a path, what to do if the turnout position changes to another position than required for the path?

I would force the position back to what is needed for the reserved path.
We need to consider if 2 rapid switching operations can damage the motor.

server/src/board/tile/rail/decouplerrailtile.cpp
gfgit: Yes but if in the future other operations are added inside `setState()` it might be worth calling them too.
`setValueInternal()` is more of an implementation design of decoupler in my opinion.

gfgit:
There is something wrong with decouplers...

gfgit:
Should "none" action always match (return true)?
Like if it's interpreted as "do nothing so whatever state is good".

server/src/board/tile/rail/decouplerrailtile.cpp
reinder: ok, that could be possible indeed, then add a private/protected `setStateInternal()` that has the bypass option. The public one shouldn't have a bypass option.

reinder:
Yeah recursion is a thing, some complex signals use up to 4 addresses where the first is a "mode address" and the others are "trigger addresses". 

> Should "none" action always match (return true)?
> Like if it's interpreted as "do nothing so whatever state is good".

Yes, that is correct.

reinder:
To make a reservation the starting block must contain a train (either occupied or reserved). The reservation system isn't fully functional yet.

> Also I've noticed that if I change the block ID, it does not update inside NX Button's block field.
> It also doesn't update block's sensor item fields block_N.input_map.items.item1.
> Here N is kept at old number.

Ok, that is something that need to be fixed.

server/src/board/tile/rail/decouplerrailtile.cpp
gfgit: `DecouplerRailTile::setState()` is already private actually. Should I rename it?

reinder:
The last page of the [DR4018 Manual](https://archive.traintastic.org/digikeijs/DR4018-Manual.pdf) manual shows some mappings. That is 8 aspects coded "binary" using the accessory addresses.

For the YaMoRC YD8116 I configured a signal with many aspects, it uses the Trigger/Mode switching sequence:
DB H-V Ausfahr + Zs2 + Zs3

I have both the DR4018 and the YD8116 so I can do some tests with it if needed :) Currently Traintastic only has two generic signals, but it would be nice to support different signaling systems in the future.

The YD8116 also support DCCext (that's why I bought one :P), that already works for the command stations supporting it.

server/src/board/tile/rail/decouplerrailtile.cpp
reinder: No, my bad, I confused it with `setPosition` of the Turnouts, that one is public because of the scripting engine.
Just leav it like this (for now at least ;))

gfgit:
Rebased on current master to fix merge conflicts

server/src/hardware/output/map/outputmap.cpp
```diff
@@ -586,18 +615,37 @@ void OutputMap::addOutput(OutputChannel ch, uint32_t id)
...
+  else if(auto* aspect = dynamic_cast<AspectOutput*>(output.get()))
+  {
+    conn = aspect->onValueChanged.connect([this](int16_t, const std::shared_ptr<AspectOutput>&){ updateStateFromOutput(); });
+  }
+  else if(auto* ecosState = dynamic_cast<ECoSStateOutput*>(output.get()))
+  {
+    conn = ecosState->onValueChanged.connect([this](uint8_t, const std::shared_ptr<ECoSStateOutput>&){ updateStateFromOutput(); });
+  }
+
+  return {output, conn};
```
reinder, line 644: Why return the connection too? Can't we just save it here in map, e.g.:
`std::map<std::shared_ptr<Output>, boost::signals2::connection> m_valueChangedConnections;` 

And use that in `releaseOutput` for releasing it.

server/src/hardware/output/map/outputmap.cpp
```diff
@@ -586,18 +615,37 @@ void OutputMap::addOutput(OutputChannel ch, uint32_t id)
...
+    conn = pair->onValueChanged.connect([this](OutputPairValue, const std::shared_ptr<PairOutput>&){ updateStateFromOutput(); });
+  }
+  else if(auto* aspect = dynamic_cast<AspectOutput*>(output.get()))
+  {
+    conn = aspect->onValueChanged.connect([this](int16_t, const std::shared_ptr<AspectOutput>&){ updateStateFromOutput(); });
+  }
+  else if(auto* ecosState = dynamic_cast<ECoSStateOutput*>(output.get()))
+  {
+    conn = ecosState->onValueChanged.connect([this](uint8_t, const std::shared_ptr<ECoSStateOutput>&){ updateStateFromOutput(); });
+  }
```
reinder, line 642: We should really get rid of this, we should add a valueChanged (without value) signal to `Output`, then that can be used to trigger `updateStateFromOutput();`

server/src/hardware/output/map/outputmap.cpp
gfgit: I think it's more efficient this way and lifetime of output reference and it's connection is the same so it makes sense to me.

server/src/hardware/output/map/outputmap.cpp
gfgit: Done! Do you like the new implementation?

server/src/hardware/output/map/outputmap.cpp
reinder: ok

reinder (commented):
Looking good, just two minor suggestions :)

reinder:
@gfgit I'd love to integrate this feature, yesterday I did some improvements on the train tracking, when this feature is integrated Traintastic should be able to track manual controlled train movement across blocks and turnouts :)

gfgit:
Hi Reinder, great work!
In this PR I'm usinge `m_keys` member of `OutputMapBase` class but you've removed in commit cdbd8a637
Should I revert it or try to find a different way?

gfgit:
Something's broken after recent commits...

gfgit:
I've rebased on master and fixed conflicts. In the process I've also removed an extra comma in single slip rail tile.
Now the logic is a bit more complex to allow matching straight A and straight B if crossed position is set to "none".
This however poses a question: should we set turnout state to Unknown if no match is found?
But we need to take in account items with multiple outputs attached. Since changing all outputs is not an atomic operation, we will get some partial state in between which is fine and should not result in "Unknown" position, if not for a brief moment.

reinder:
Nice work! I'll try to take look into it this evening. :)

server/src/hardware/output/output.cpp
```diff
@@ -35,11 +35,14 @@ Output::Output(std::shared_ptr<OutputController> outputController, OutputChannel
   : interface{this, "interface", std::move(outputController), PropertyFlags::Constant | PropertyFlags::NoStore | PropertyFlags::ScriptReadOnly}
   , channel{this, "channel", channel_, PropertyFlags::Constant | PropertyFlags::NoStore | PropertyFlags::ScriptReadOnly}
   , type{this, "type", type_, PropertyFlags::Constant | PropertyFlags::NoStore | PropertyFlags::ScriptReadOnly}
+  , onValueChangedGeneric{*this, "on_value_changed_generic", EventFlags::Scriptable}
```
reinder, line 38: `EventFlags::Scriptable` can be removed, in Lua the `onValueChanged` events can be used, in Lua you can use the same handler for different output's as Lua has weak typing.

Changing this will also fix the failing "Lua manual" build :)

gfgit: Done!

reinder:
Did some tests in simulation mode, it is working great :)

Server release builds are failing, I'd guess it is a missing `method.tpp` include.

I'd hope I can do a real world tests soon, once it builds I'll merge it :)

Issue #114, opened 2024-04-19
Allow multiple compatible reservation


gfgit:
**Describe the bug**
Some path reservation are compatible to each other, meaning they both can be active at same time.
But they could share some tiles like a bridge which can allow both trains to pass above and below, as opposed to a crossing which allows only one train at a time (and so only one reserved path).

So for bridges we to allow multiple reservations otherwise when releasing first path, the bridge will look as released (reserve state is zero) but the other path reservation is still active.
I suggest a system similar to reference-counting. When it's zero the object is free (es. a turnout can be switched) otherwise if reference count is greater than zero the object is locked (es turnout locked in reserved position).

**Future expansion**
In the future this could be expanded to allow moving turnouts not directly traveled by train in path but needed to ensure no other train would collide with it.

 
![Screenshot_20240419_155607](https://github.com/traintastic/traintastic/assets/42845724/c4de1621-6854-41db-8056-a6f9aecf83c4)

At least for Italian railways the situation in the picture above would be wrong.
The upper turnout which currently is in "Left" position should be in "Right" position (straight).
This ensures that if a train accidentally departs from "block_13" it will not collide with our train entering (or exiting) the station trough the reserved path.
Instead a train coming from "block_13" would go to the dead track ("block_16" in this case).

gfgit:
Currently I've implemented:
- Cancelling a reserved path by pressing on same NX buttons which were used to reserve it. (Currently works but bypasses some checks which need to be reworked)
- Prevent 2 paths to reserve same turnout for different positions (this was a bug in `reserve()` method).
- Locking turnout position if it's in a reserved path.
- React to turnout external changes and revert them if in reserved path.

reinder:
Bridge reservation currently works with a bit mask, both paths have their own bit, see: https://github.com/traintastic/traintastic/blob/master/server/src/board/tile/rail/bridgerailtile.cpp

Regarding the "Italian" way of turnout control, easiest would be to just use the same address for both turnouts. (Or somehow link them if the both have a different address,) Then the user can easily decide how they want it to work.

Other option could be to do it automagically based on the used signaling system.

gfgit:
> Bridge reservation currently works with a bit mask, both paths have their own bit

Hm, it didn't work when I've tested it. I'll test again later

reinder:
We should add some tests to check that, also the other things you mentioned for turnout control while it is reserved.

reinder:
[...]

PR #115, opened 2024-04-19
Path reservation fixes

gfgit:
Some experiments with path reservation system.
Based on #113 

Some issues already noted in #114


server/src/board/tile/rail/directioncontrolrailtile.cpp
```diff
@@ -61,6 +61,8 @@ DirectionControlRailTile::DirectionControlRailTile(World& world, std::string_vie
   , setState{*this, "set_state", MethodFlags::ScriptCallable,
       [this](DirectionControlState newState)
       {
+        if(reservedState())
+          return false; // Direction control is currently locked by reserved path
```
reinder, line 65: Just a thought:
If we have `A--(<>)--B` (=both) and there is a reserved path from A to B, should we allow changing it from *both* to *A -> B* ? or the other way around.

As long as it doesn't conflict with the reservation should we allow changing it?

gfgit: Fully agree. Done!

reinder: Awesome!

reinder:
Maybe we can add some tests to check e.g. reservation and changing state of a direction tile while it is reserved.

reinder:
I created a small test for the direction tile state change while it is reserverd, can you apply the patch to you branch?
[0001-test-added-Direction-path-reservation-using-NX-and-c.zip](https://github.com/traintastic/traintastic/files/15083099/0001-test-added-Direction-path-reservation-using-NX-and-c.zip)

I zipped it because GitHub refuses to attach the PATCH file...but the error explains that PATCH is supported 🤷🏼‍♂️

gfgit:
Hi, I've applied your patch.
Also I've noticed sometimes NX Buttons work even if block rail tile is not adjacent, like `NX Button + straight rail + Block tile`.
Sometimes it does not work. But when I read the `findBlock()` code I think it is supposed to never work.

reinder:
Awesome.

It is indeed required to be adjacent to a block but, passive tiles are ignored. On the `boardModified` event a new graph network is built of the board. It only creates nodes for tiles that have a function, passive tiles like straight en curved rail pieces are ignored. So for a NX button there may be straight/curved rail pieces between them, but not something like a signal/turnout/direction control etc.

gfgit:
What about bridges? In my test board it didn't work if a bridge was inbetween NX button and block tile.

PS. Let's see ho replying by mail works out...

Il 25 aprile 2024 19:39:20 CEST, Reinder Feenstra ***@***.***> ha scritto:

reinder:
A bridge is a node (due to two seperate reservation paths), so that's why it doesn't work, That is a bug, a bridge is a passive tile so it should be allowed. Going to fix that.

reinder:
If you merge/rebase latest `master` the bridge nx button issue is fixed :)

gfgit:
Rebased on current master.
Some commits are really trivial fixes which could be merged already, while others contain more complex changes on which we should discuss.
If you want I can split in 2 distinct PRs

gfgit:
Fixed missing include in `blockpath.cpp`
Also now trivial changes are moved in first commits while last commits contain complex changes

client/src/board/boardwidget.cpp
```diff
@@ -778,6 +768,61 @@ void BoardWidget::rightClicked()
     rotateTile(true);
 }
 
+void BoardWidget::startHoldTimer(const std::shared_ptr<NXButtonRailTile>& nxButton)
+{
+  stopTimerAndReleaseButtons();
+
+  m_releaseButton1 = nxButton;
+
+  m_timer = new QTimer(this);
```
reinder, line 777: I suggest to add a `assert(m_timer);` here, to check if `m_timer` is `nullptr` as expected.

client/src/board/boardwidget.cpp
```diff
@@ -778,6 +768,61 @@ void BoardWidget::rightClicked()
...
+{
+  // Do not release first button yet
+  m_releaseButton1.reset();
+
+  stopTimerAndReleaseButtons();
+
+  m_releaseButton1 = firstButton;
+  m_releaseButton2 = nxButton;
+
+  m_timer = new QTimer(this);
```
reinder, line 796: I suggest to add a `assert(m_timer);` here, to check if `m_timer` is `nullptr` as expected.

reinder:
Nice improvements, I'm going to do some test with it :)

gfgit:
Hi Reinder. Yes I should add an assert. But if you test the code and it works I could change this to avoud using QTimer which requires a new heap allocation every time we start it and instead use `QTimerEvent` and `startTimer()` `killTimer()`

Functionality is the same and assert would become timer ID equal to zero.

Il 11 giugno 2024 23:09:08 CEST, Reinder Feenstra ***@***.***> ha scritto:

reinder:
Good suggenstion, `QTimerEvent` :)

Just did some simulations tests, it works as expected :)

I was thinking, should we log a warning if a turnout is modified externally an Traintastic corrects it? Should we also have some kind of max retry to prevent loops? and maybe Estop the train if the max retry count is reached and log an error then too?

For the path release, there must be some kind of delay I think, the train can still be on the turnout but already left the tail block. (Ideally we have occupy sensors on the turnout's too offcourse, but Traintastic should be able to handle both situations.)

gfgit:
Yes we should log a warning when turnout is externally changed. But I don't know how to add a new log message...
As for the max retry count when it's reached maybe yes we E-stop the train. Although this is a strong action which might not be good for some users. Maybe we add a setting like "E-stop train if on path and turnouts  are changed externally".

I think I've added the retry count logic to signals, if it's working fine I'll copy it also to turnouts.

For the path release, I bet very few layouts have dedicated occupy detectors for switches so yes we should workaround it with some kind of delay.
Better yet if we know the real train speed and it's real length we can estimate traveled space and so have a varying delay according to train speed and length plus a security margin.
We don't even have to know path length which would depend on turnout state in case of multiple possible paths for the same pair of NX buttons. We start to count from when exit block gets occupied, until train tail is inside the exit block and entrance block is free.

reinder:
> Yes we should log a warning when turnout is externally changed. But I don't know how to add a new log message...
I'll explain :)

To add a log message you need to add a enum value in `[/shared/src/traintastic/enum/logmessage.hpp](https://github.com/traintastic/traintastic/blob/master/shared/src/traintastic/enum/logmessage.hpp)`.
In this case it is a train/board related message so it is in the `3xxx` range, just use the next free number.
You also need to add at least an English term for the message in, as the server always uses English for the console log.

I've numbered all the log messages so it is easy to look up, in the source and if we receive a screenshot/log text in a language we don't understand :)

> As for the max retry count when it's reached maybe yes we E-stop the train. Although this is a strong action which might not be good for some users. Maybe we add a setting like "E-stop train if on path and turnouts are changed externally".

Good suggestion to add a world setting for that, maybe we can add three options:
- Ignore
- Estop train (= default?)
- Stop world

> I think I've added the retry count logic to signals, if it's working fine I'll copy it also to turnouts.

I didn't test that yet, but it seems fine, else we can improve it later :)
 
> For the path release, I bet very few layouts have dedicated occupy detectors for switches so yes we should workaround it with some kind of delay. Better yet if we know the real train speed and it's real length we can estimate traveled space and so have a varying delay according to train speed and length plus a security margin. We don't even have to know path length which would depend on turnout state in case of multiple possible paths for the same pair of NX buttons. We start to count from when exit block gets occupied, until train tail is inside the exit block and entrance block is free.

Good suggestions/ideas! I think it is best we move that to a separate PR, then we can finish this one :)

reinder:
Nice improvements!

Just some thoughts:
- The log message is *Position externally changed while locked in a path* shouldn't we add a separate message for the signal? (A signal has an *aspect* not a *position*.) Should it also be mentioned that Traintastic is going to correct it?
- When a Train is E-stopped, I think a error must be logged to explain why the train is E-stopped.

gfgit:
> * When a Train is E-stopped, I think a error must be logged to explain why the train is E-stopped.

Sure. I was thinking about moving the "stop all trains" code inside `BlockPath` since it's duplicated in Turnout and Signal.
But then the log output cannot know the reason it's been called. Will think more about it.

gfgit:
Last things missing should be:
- Opt-out E-Stopping trains in World settings (should this be a `Property<bool>` inside World object?)
- Maybe also stop the whole World
- Delay for releasing path after train has freed entrance block

reinder:
> Last things missing should be:
> 
>     * Opt-out E-Stopping trains in World settings (should this be a `Property<bool>` inside World object?)
>     * Maybe also stop the whole World

Yes, you can add a `Property<>` to the `World` for that, it will automagically appear in the World dialog, there aren't much settings yet, but for the future it is best to add it to the *Trains* category, as examples see `server/src/traintastic/settings.cpp` (The UI then will create a *Trains* tab with the setting.)

I think stop the whole world is also a nice option to have, so then we need an enum.

>     * Delay for releasing path after train has freed entrance block

That would be nice indeed.

gfgit:
I've added World settings and Italian translations.
Item names are a bit too long but couldn't think of shorter names...

How do I add a delay for path release?
It seems I must access IO context of Eventloop which is private

reinder:
> I've added World settings and Italian translations. Item names are a bit too long but couldn't think of shorter names... 

Awesome! Above all it must be understandable for the user, so if that requires a longer text than it is fine :)

> How do I add a delay for path release? It seems I must access IO context of Eventloop which is private

It's `public`, the `Clock` also uses the `EventLoop::ioContext` for its `boost::asio::steady_timer`.

gfgit:
> It's `public`, the `Clock` also uses the `EventLoop::ioContext` for its `boost::asio::steady_timer`.

Done and rebased on master.

There is one issue with current implementation:
If train is so short that it will free entrance block before occupying exit block, the logic will not release the path.
Also simulating re-occupying entrance block will not work because entrance block does not belong to the Train anymore.
So the path will freeze in reserved state until it's manually released by user clicking same buttons in same order as for reserving it.
This is caused by the condition "keep train in at least one block else we loose it" which is correct but does not play well when there is a section of track without sensors between 2 consecutive blocks.

Anther thing that came to my mind is that in case of high traffic it may be good to let user reserve path in advance, even before Train occupies entrance block.
Otherwise the incoming Train always sees a "yellow" (proceed but stop at next signal) and a "red" (stop).
Then the block is occupied so station master can reserve a path.
Signal changes to "proceed" and Train starts moving again and enters the station.

Last week I've visited a station control panel in operation.
There was an incoming freight train, an alarm rang when train was approaching the station but still far from it, at least 2 blocks far. So train had 2 signals ahead and then station's entrance signal.
`Train -> sig1 (green) -> block1 (free) -> sig2 (yellow) -> block2 (free) -> Station Entrance sig (red) -> Station`.

The station master reacted to the alarm by reserving a path to let train enter on platform 1.
This operation quit the alarm and set signals to green:

`Train -> sig1 (green) -> block1 (free) -> sig2 (green) -> block2 (free) -> Station Entrance sig (green) -> Station (path to platf. 1, pass through)`.

The train driver never saw the yellow and red signals because they were changed in time before train reached them, so train didn't have to stop before entering the station.

reinder:
> > It's `public`, the `Clock` also uses the `EventLoop::ioContext` for its `boost::asio::steady_timer`.
> 
> Done and rebased on master.

Nice work!

> There is one issue with current implementation: If train is so short that it will free entrance block before occupying exit block, the logic will not release the path. Also simulating re-occupying entrance block will not work because entrance block does not belong to the Train anymore. So the path will freeze in reserved state until it's manually released by user clicking same buttons in same order as for reserving it. This is caused by the condition "keep train in at least one block else we loose it" which is correct but does not play well when there is a section of track without sensors between 2 consecutive blocks.

Yeah I just discovered that too while working on the #144 zone support, I recently made a fix (https://github.com/traintastic/traintastic/commit/c18dcb570e247cc540046379011c51a02d2f474d) so tailing reserved blocks are released once the train is occupying a block again, that was a similar issue to this one. We can rename `fireOnTrainLeft` to e.g. `trainLeft` then it can release the path and the fire the event.

> Anther thing that came to my mind is that in case of high traffic it may be good to let user reserve path in advance, even before Train occupies entrance block. Otherwise the incoming Train always sees a "yellow" (proceed but stop at next signal) and a "red" (stop). Then the block is occupied so station master can reserve a path. Signal changes to "proceed" and Train starts moving again and enters the station.

A train could even reserve all block until the next turnout, as there won't be another train there anyway. We also can add a setting for that, how many block a train should try to reserve ahead. Maybe we can discuss that a bit in the discussion section and finish this PR :)

> Last week I've visited a station control panel in operation. There was an incoming freight train, an alarm rang when train was approaching the station but still far from it, at least 2 blocks far. So train had 2 signals ahead and then station's entrance signal. `Train -> sig1 (green) -> block1 (free) -> sig2 (yellow) -> block2 (free) -> Station Entrance sig (red) -> Station`.
> 
> The station master reacted to the alarm by reserving a path to let train enter on platform 1. This operation quit the alarm and set signals to green:
> 
> `Train -> sig1 (green) -> block1 (free) -> sig2 (green) -> block2 (free) -> Station Entrance sig (green) -> Station (path to platf. 1, pass through)`.
> 
> The train driver never saw the yellow and red signals because they were changed in time before train reached them, so train didn't have to stop before entering the station.

Awesome, were you allowed to make any photo's there? It would be awesome if Traintastic could do that in the future too. Might require some Lua scripting :) Did you see the zone feature (#144)?

gfgit:
Hi. Zone feature seems really cool!
I saw your commit to mitigare short trains but I only tested before rebasing so cannot say if it fixes the issue. I will not be able to work on this for this and next week.

Reaerving ahead until next turnout is at same time cool and akward. I agree to discuss it separately after this PR.

At the station unfortunately I was not able to take any photos. I could try to ask again but I don't know if then I could post them publicly here.
In case I could make a schematic of the real situation to circumvent this birocratic limitation.
I'll let you know!

reinder:
> Hi. Zone feature seems really cool!

Yeah, I expect it to give many posibilities in the future, I won't include it in the 0.3 release so we have time to discuss it further for use cases. Currently the is only one Zone type, we could add different types so their behavoiur can be optimized to a specific use case, e.g. shadow station, shunting area or station. Other option is having a single zone type with many options. Both have their pro's and cons, feel free to post any comment to the issue :) 

> I saw your commit to mitigare short trains but I only tested before rebasing so cannot say if it fixes the issue.

It won't fix it, the path release is triggerd by the sensors, not by the *train left* event. I think we should change that. Maybe even delay the block release instead of the path release.

> I will not be able to work on this for this and next week. 

No problem, thanks for letting me know, I really like to include this PR in Traintastic v0.3 (no pressure :wink:). The 0.2 release is more that a year ago and during that time many things have improved. Once 0.3 is released I'd like to discuss the goal for 0.4, hopefully some other contributers will join too.

> Reaerving ahead until next turnout is at same time cool and akward. I agree to discuss it separately after this PR. 

:+1: 

> At the station unfortunately I was not able to take any photos. I could try to ask again but I don't know if then I could post them publicly here. In case I could make a schematic of the real situation to circumvent this birocratic limitation. I'll let you know!

Posting that kind of info publicly can indeed be an issue, always good to check with them first :) I was curious because in the very future it would be awesome if the board system could model such a system just like the real world.

reinder:
That is indeed nice, I renumbered mine :)

gfgit:
Thx. I've fixed the short train issue. Will rebase-push soon

reinder:
Ok, great, no hurry, can't run the pipeline at the moment, the traintastic repo exceeded the LFS bandwidth by 150% #oops

gfgit:
> Ok, great, no hurry, can't run the pipeline at the moment, the traintastic repo exceeded the LFS bandwidth by 150% #oops

Uh. What's that?

reinder:
@gfgit I stored a few windows DLL files in git LFS, for every build job they are downloaded. But that bandwidth is metered, for an organization the limit is 1GB/month. I'm going to move those file to "normal" git, that isn't metered :)

reinder:
I just removed the git lfs stuff, build is now running :) If that works can you merge/rebase `master`?

gfgit:
I've rebased on master.

reinder:
@gfgit can I consider this one *ready for review*?

gfgit:
Hi. Yes it's ready!

reinder (commented):
Looking good! just a few notes, can you have a look?

gfgit:
Hi, this week and next one I'll be busy. I agree on preposed changes, maybe end of next week I can make them.

reinder:
Awesome 😎 thanks for the response :) good luck with the busy stuff.

server/src/board/tile/rail/turnout/turnoutrailtile.cpp
```diff
@@ -54,19 +65,32 @@ TurnoutRailTile::TurnoutRailTile(World& world, std::string_view _id, TileId tile
...
   if(!isValidPosition(turnoutPosition))
   {
     return false;
   }
+
+  const TurnoutPosition reservedPos = getReservedPosition();
+  if(reservedPos != TurnoutPosition::Unknown && reservedPos != turnoutPosition)
+  {
+    // TODO: what if 2 path reserve same turnout for same position?
+    // Upon release one path it will make turnout free while it's still reserved by second path
```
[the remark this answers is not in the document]
gfgit: Should I remove comment?

shared/src/traintastic/enum/externaloutputchangeaction.hpp
```diff
@@ -0,0 +1,31 @@
+#ifndef TRAINTASTIC_SHARED_TRAINTASTIC_ENUM_EXTERNALOUTPUTCHANGEACTION_HPP
```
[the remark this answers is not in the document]
gfgit: Done

gfgit:
@reinder 
While resolving conflicts I've noticed this:

```
void TurnoutRailTile::destroying()
{
  outputMap->parentObject.setValueInternal(nullptr);
  RailTile::addToWorld();
}
```

Shouldn't it call `RailTile::destroying();` instead? There are multiple situations like this made from same commit.

reinder:
> @reinder While resolving conflicts I've noticed this:
> 
> ```
> void TurnoutRailTile::destroying()
> {
>   outputMap->parentObject.setValueInternal(nullptr);
>   RailTile::addToWorld();
> }
> ```
> 
> Shouldn't it call `RailTile::destroying();` instead? There are multiple situations like this made from same commit.

Yes! that is very wrong indeed, nice catch. :ok_hand:

server/src/board/tile/rail/turnout/turnoutrailtile.cpp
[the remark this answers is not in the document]
reinder: Just leave it for later, I just wanted to explain that that isn't a possibility currently.

PR #119, opened 2024-04-29
WIP: Italian signals

gfgit:
Hi, I've made a really quick'n dirty patch to add Italian signalling system.
It was really fun to code!


reinder:
Nice work, I didn't have the time to view all the changes yet. The blinking option is really nice :)

But I was thinking, the signal aspects, there are in fact two options:
- Use one `SignalAspect` enum for all signals or
- use give each signal its own enum.

I'd expect that there is a certain overlap between signal aspects for different signaling systems, at least *Stop* overlaps :P

Based on a quick scan of the code, if I understand it right, the ITA signal has three lamps that are all multi color, is that correct?

gfgit:
> But I was thinking, the signal aspects, there are in fact two options:
> 
>     * Use one `SignalAspect` enum for all signals or
> 
>     * use give each signal its own enum.

You've anticipated me!
I've implemented a mixed approach.
`SignalAspectITA` is cast to basic `SignalAspect` upon setting it through normal `setAspect` method.
Inside it's converted to a basic `SignalAspect` which goes to `aspect` property for interoperability with 2/3 aspects signals (say one wants to reproduce railways  crossing national borders) and the "full" `SignalAspectITA` value is stored in a mirroring property `aspectITA` which is used only for drawing tile and by other Italian signals in the same signal path.

> I'd expect that there is a certain overlap between signal aspects for different signaling systems, at least _Stop_ overlaps :P

Yes, in fact first 4 enum values of `SignalAspectITA` have same meaning and **same value** of basic `SignalAspect` so they could be cast without conversion.

> Based on a quick scan of the code, if I understand it right, the ITA signal has three lamps that are all multi color, is that correct?

Italian signals can have up to 3 multi color lights. Only the top most light can show red. On railway lines typically you find signals with 1 or 2 lights which can show a subset of existing aspects, while on junctions and stations you can also find 3 light signals which can show all aspects based on layout necessities.
There are also other less important aspects and signal types which I have not yet implemented :)

There is also the concept of anticipating the aspect ("Avviso" in Italian) in the previous encountered signal.
So for example if a first category signal imposes 30 km/h speed limit because of some short radius turnout in non-straight position, then the previous signal shows a "proceed with no speed restrictions but then slow down to reach 30 km/h before next signal".

Then they invented joined "anticipation + first category" signals which can at the same time impose a speed limit from the signal itself (anticipated by the previous signal) and anticipate next signal speed limit (or other aspects) so driver can adjust in time before reaching next signal.

This is why I've introduced `SignalAspectITA_ingredients` which is a binary OR'ed flag. The most important aspects made by combining the ingredients are added to `SignalAspectITA` for convenience. Not all combinations are valid, so I will code some logic to reject them.

Light state and color is then calculated and stored. This makes it easier to draw on client side but makes it impossible to draw every aspect on demand for the tile context menu. Need to rethink this a bit...

reinder:
Thanks for the explanation, now I understand it al lot better.

So depending on the aspects needed there are one, two or three lamp signals. Would it then make sense to add three separate tiles for that? (Or 6 for with or without pre signal support.)
For the end user that is easier I think, then the number of aspect can be limited due to the number of lamps.

The border thing is indeed an thing to take into account. We could however choose to only support real world border.s Italy <-> The Netherlands is not something that needs an implementation. As fallback we can also implement a few functions that can help in those cases, e.g. `isStop(SignalAspect)` or `isSpeedRestriction(SignalAspect)` etc.

In Germany there are also combined pre/main signals, but they both have their own lamps so that is easier. That can also be implemented by using two separate tiles. So I was thinking, what if we add an additional property for the pre signal aspect, e.g. `preAspect`? That will have an impact on the output map however, so we need to check if that doesn't make it harder.

> Light state and color is then calculated and stored. This makes it easier to draw on client side but makes it impossible to draw every aspect on demand for the tile context menu. Need to rethink this a bit...

I prefer to move that kind of logic to the client, the server doesn't care about the lamp states etc. it only needs the aspect :)
And indeed for drawing the signal in the context menu (and output map #todo) it is needed anyway.

To find out if creating one Enum will all signal aspects is a good solution we should create a list of aspects for various signaling systems. I expect there is some overlap. Roco has for their Z21 signal decoder a list of aspects, see: https://www.z21.eu/en/products/z21-signal-decoder/signaltypen (in German), they have not support for Italian :(

My current idea for the client is signal menu is to create another level of menus:
- Generic
  - Two aspect signal
  - Three aspect signal
- Germany H/V
  - ...
- Germany Ks
  - ...
- Germany Hl
  - ...
- Italy
  - One lamp
  - Two lamps
  - Three lamps
- The Netherlands
  - Main signal
  - Dwarf signal
etc.

And also use this structure for the board tile folders.

gfgit:
Making a list is a good idea!
About multiple tiles, wouldn't it be better to just add a property numberOfLamps which will change the drawing of tile and reject aspects which need more lamps?

What are Netherlands dwarf signals? We also have dwarf signals in Italy but they are used only in stations and only show "stop" or "proceed".
They tell you that turnouts are set correctly so they are a pre-condition to set main signal to "proceed" or any other non-stop aspects.
For shunter locomotives which just move inside the station the main signal stays red because you cannot go to main line but dwarf signals are set to "proceed" accordingly to turnout positions.
Italian dwarf signals must go on a separate tile because you can have multiple of them in a path between 2 nx buttons while you have only a single main signal.

Il 30 aprile 2024 22:13:28 CEST, Reinder Feenstra ***@***.***> ha scritto:

reinder:
> Making a list is a good idea!

Some spreadsheet like document would work I think, do you have a suggestion for a privacy friendly one?

> About multiple tiles, wouldn't it be better to just add a property numberOfLamps which will change the drawing of tile and reject aspects which need more lamps?

Fair point. Currently the list of aspects in the output map is fixed (once set on construction), so that will require a bit of a refactor then. Signals with more aspects probably have properties to control their behavior, so it might reduce complexity to separate that, for some signals however it might introduce the risk of code duplication.

We could also build one signal tile that has a `type` property in which you can select: Generic 2 aspect / Generic 3 aspect / NL main / NL dwarf / ITA main / ITA dwarf etc. Signal logic and settings could then be moved to a sub object (similar to `interface.settings`).

> What are Netherlands dwarf signals?

Images of dutch dwarf signals (dwergsein): https://duckduckgo.com/?q=dwergsein&t=ffab&iax=images&ia=images

They have 4 aspects:
- Red: Stop
- Yellow blinking: Proceed with max. 40 km/h and be able to stop at any time.
- Yellow: Proceed with max. 40 km/h, expect Stop
- Green: Proceed with max. 40 km/h

They are usually used in and near stations. Yellow blinking is used when a train is entering an already occupied block or entering an unprotected area.

Apart from their size and speed limit they are similar to the main signals.

> Italian dwarf signals must go on a separate tile because you can have multiple of them in a path between 2 nx buttons while you have only a single main signal.

That's interesting, totally different than the dutch dwarf signal. I'd expect the whole signalling thing will be a challenge with all these different signalling systems :)

gfgit:
About the spreadsheet I can make one in my nexcloud and share a public link.
Then once it's complete you can move it in your website maybe in the form of a wiki page.

Instead of extracting an "engine" which controls a generic multipurpose signal tile which could be difficult to do, we could instead do a base class for each signal system and derive it.
Like Italian base system has the logic and subclasses set the allowed aspects in constructor.
Lime it's currently done for turnouts.
Also signals of same system should be "upgradable" by dropping a tile onto another


Il 1 maggio 2024 23:14:52 CEST, Reinder Feenstra ***@***.***> ha scritto:

reinder:
> About the spreadsheet I can make one in my nexcloud and share a public link.

Awesome, can you send me the link. maybe @princeofnaxos can add some details about the Swedish signaling system.

> Then once it's complete you can move it in your website maybe in the form of a wiki page. 

Or we could convert it to markdown an add it to git, e.g. in a `docs` folder. Then it is easier too keep it in sync with the code.

> Instead of extracting an "engine" which controls a generic multipurpose signal tile which could be difficult to do, we could instead do a base class for each signal system and derive it. Like Italian base system has the logic and subclasses set the allowed aspects in constructor. Lime it's currently done for turnouts. Also signals of same system should be "upgradable" by dropping a tile onto another.

The benefit of using a generic multipurpose signal tile would be that changing/upgrading the signal type is easy, just select another `type`. However when using separate tiles it is easier to support a limited number of upgrades. e.g. disallow change from ITA to NL.

Currently "sub object" are kind of const, these properties are set once at construction but never change afterwards. (The object the property refers to isn't const of course.)

Maybe we can experiment with both and see what works best for developers and end users. The number of signaling systems could easily reach 20 I think. Changing object properties at runtime will also require client modifications as that isn't supported currently. I'm curious if it is a good or a bad idea, so I'll create a branch to do some experiments next week.

p.s. I like to release Traintastic 0.3 soon, mainly to have a new release 0.2 is over a year ago, so this signaling stuff will not go into 0.3. For 0.4 I'd like to make Traintasic have the base minimum features for block / some form of automatic trains. Signaling would be a good addition, as the different systems might give insights on things to take into account for automatic trains.

reinder:
> We can try this: https://ethercalc.net/=x3pz4ie5o997

I added the generic and NL signals and saved a local copy (as anyone reading this can edit it now :sunglasses:)

> I did't understand the sub object thing... 

My bad, I'd hope to build a small demo next week on a branch to demonstrate and try to explain it again :)

> About automation we should add a "lengt" property to blocks to calculate brake distance.

And you need block / turnout position speed limits to determine the correct aspect.

> Also add a generic way to extract speed limit from any type of signal.

Yeah, for sure. While adding the NL aspects, each aspect is basically a combination of a few things. For the ITA signals it is the same I think, I'd hope there is some overlap :)

gfgit:
@reinder I saw you added NL signals. I like this table approach.
But... I think there might be a bug in ethercalc when saving :(
If I download the file locally in ODS or EXCEL format it's full of garbage, probably wrong character encoding.
Only HTML and CSV seems to be working.

I've updated a new file at: https://ethercalc.net/=railwaysignals
which has the correct sheet name and some initial Italian signals integrated in your table scheme

reinder:
Yeah, there is indeed a bug in Ethercalc, the ODS I  downloaded is garbage too.

Nice work, there are really a lot of aspects!

> Proceed, max 30 km/h, brake until 30 km/h, must be reached at next signal

I don't get this one.
- Is the speed limit 30 km/h? (then there is no need to brake) or
- Is it allowed to pass at a higher speed but must the train slowdown until 30 km/h?

I added a few column in the front (not yet complete), I thought it would be nice if we can break down the aspects into "attributes", we can use that more easily for train control.

gfgit:
> * Is the speed limit 30 km/h? (then there is no need to brake) or

Yea sorry, I've just copied "brake" from the line above. The meaning is:

- You can pass at 30 km/h
- Next signal will allow passing at 30 km/h

So basically you're always at 30 km/h until maybe next's next signal :)

gfgit:
Instead what  does it mean: "Proceed, brake until 60 km/h, not required to be reached at next signal"?
I can pass the signal at any speed and then brake until 60 km/h even after next signal?

reinder:
> Instead what does it mean: "Proceed, brake until 60 km/h, not required to be reached at next signal"? I can pass the signal at any speed and then brake until 60 km/h even after next signal?

That is correct, the next signal then will be "brake until 60km/h must be reached at next signal". I think it is a short block thing, we're one block isn't enough to brake to 60 km/h.

gfgit:
By the way we also have "candle tree" signals (In Italian "Segnali a candeliere") which unfortunately were replaced around 2017 so I've never got to see one in action.
They were used in junctions, station entrances and station departure signal to tell driver which direction will be taken by the train.
This is one example: Segnale a candelabro
But there are many combination, for example in symmetric junction there is not central topmost lamp.
Or it's only left or only right like this: Segnale partenza stazione di Perugia

gfgit:
Hi @reinder, I'm trying to understand Block reservation logic.
They have a `ObjectVectorProperty<TrainBlockStatus> trains` property but it seems even if multiple inputs are attached to a block, only one train can be assigned to it.
While this is correct for normal mid-line blocks, in stations it should be allowed to let 2 trains enter the same track, provided that block has at least 2 inputs, first train only occupies second input and is stopped, then following train can enter with reduced speed and stop at the end of block's first input limit.
We could add a new boolean property "allowMultipleTrains" which defaults to false and is disabled if there is a single input.
I can remove some of the "return false" inside the `BlockTile::reserve()` but I want to understand if there could be side effects first.
Also how should client draw block tile if there are multiple trains in it?
This scenario is perfect for merging 2 passenger trains in a station which then go to a common destination.

reinder:
> I'm trying to understand Block reservation logic. They have a `ObjectVectorProperty<TrainBlockStatus> trains` property but it seems even if multiple inputs are attached to a block, only one train can be assigned to it. While this is correct for normal mid-line blocks, in stations it should be allowed to let 2 trains enter the same track, provided that block has at least 2 inputs, first train only occupies second input and is stopped, then following train can enter with reduced speed and stop at the end of block's first input limit.

That is indeed not supported yet, the `trains` property is already a vector so it can be supported in the future.

> We could add a new boolean property "allowMultipleTrains" which defaults to false and is disabled if there is a single input. I can remove some of the "return false" inside the `BlockTile::reserve()` but I want to understand if there could be side effects first.

I think it is best that "normal" reservations always fails if a block is occupied. Entering an occupied block is something special with a few use cases which involve other operations too. So I think it is best to handle it as a special thing.
The property `allowMultipleTrains` is a good idea :)

> Also how should client draw block tile if there are multiple trains in it? This scenario is perfect for merging 2 passenger trains in a station which then go to a common destination.

That isn't supported yet, but should be supported in the future.

Merging trains really is an advanced operation, something we must take into account, however I think we should first implement the "normal" reservation and train tracking stuff :)

gfgit:
> That isn't supported yet, but should be supported in the future.
> 
> Merging trains really is an advanced operation, something we must take into account, however I think we should first implement the "normal" reservation and train tracking stuff :)

Sure. In the meantime I've added support for "partially occupied aspect even though it cannot be disabled yet for mainline if it detects multiple inputs for a block. When we'll move to advanced stuff it will be ready.

I've also added "anticipation-only signals" (Avviso) using same base class.
Now things todo:

1. Add tests and properly check edge cases
2. Support 1/2 lamps signals by reducing allowed aspects
3. Fix output map which crashes on save because of enum type mismatch
4. For completeness I could add "blinking yellow" which is for reduced block length but it would be rarely used in model layouts so low priority.

gfgit:
Hi @reinder I've implemented setting signal lamp count. However in client board drawing code signals with 1 lamp look ugly.
They should have a white circle around the lamp and instead the are 2 semi-arcs which intersect each other.
I've tried to keep code generic and easy to understand but now it got complicated I think...

princeofnaxos:
@reinder since this is WIP and becoming a very long conversation, I think we maybe should invite @gfgit to Pumble? Certain things should really be written here for the world to read, but many things could remain internally.

reinder:
Just created a discussion thread, as this is more about signaling in general now -> https://github.com/traintastic/traintastic/discussions/123

gfgit:
> I think we maybe should invite @gfgit to Pumble?

Hi, thanks for the invitation but I don't like that platform. I don't think it's so bad to discuss this in #123

PR #131, opened 2024-06-12
Fix z21 power state

### LINKED ISSUE #85: DR5000 with Z21 no power on is possible
power off is ok

gfgit:
This fixes power state changes being ignored from both end at times or annoying repeated changes in loop.

By the way, I've finally succeeded in making myself a small Z21 with an ESP32 and using Philipp Gahtow libraries ([available here](https://www.pgahtow.de/w/Z21_mobile/en)) with some patches on top to make it work better and as an XpressNet slave.

ESP32_Z21_mia

I cannot test with a real Z21 but you maybe could test with Digikeijs DR5000.

Should fix #85

### DIFF
--- a/server/src/hardware/interface/z21interface.cpp
+++ b/server/src/hardware/interface/z21interface.cpp
@@ -209,22 +209,41 @@ bool Z21Interface::setOnline(bool& value, bool simulation)
           else
             firmwareVersion.setValueInternal("");
         });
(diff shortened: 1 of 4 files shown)


reinder:
Sorry for the delay, I thought there was a build issue, but it was just a failing Qt install. Merging it :rocket:

gfgit:
> Sorry for the delay

No problem. Thanks!

PR #139, opened 2024-06-16
Update FindLua.cmake

DrEVILish:
/opt/homebrew # MacOS Apple Silicone


DrEVILish:
after downloading a fresh copy of the git repo and trying to build again, 
/opt/homebrew # MacOS Apple Silicone
was missing from the FindLua.cmake file.

PR #147, opened 2024-06-21
Fixed new line character skipping

DrEVILish:
\n\r handles in both directions as well \r\n as well as just \n or \r

Got stuck in a loop after reading the new line character it didn't count it in the loop, so ignored it and kept looping.


reinder:
Awesome, great work @DrEVILish! I'll merge it :rocket: 

> A line ends with either a CR (13=0x0D) or LF (10=0x0A).
> source: https://loconetovertcp.sourceforge.net/Protocol/LoconetOverTcp.html#syntax

I've implemented it according to the specification which states it is OR not both, but being a bit more relaxed won't hurt :)

PR #162, opened 2024-08-16
Unlatch momentary functions after specified timeout

gfgit:
Second commit is an adaption for Z21 ClientKernel. It should be done also for other interfaces.

See #161

### DIFF
--- a/server/src/hardware/decoder/decoder.cpp
+++ b/server/src/hardware/decoder/decoder.cpp
@@ -29,6 +29,7 @@
 #include "../protocol/dcc/dcc.hpp"
 #include "../throttle/throttle.hpp"
 #include "../../world/world.hpp"
+#include "../../core/eventloop.hpp"
 #include "../../core/objectproperty.tpp"
 #include "../../core/attributes.hpp"
 #include "../../core/method.tpp"
@@ -42,6 +43,8 @@ const std::shared_ptr<Decoder> Decoder::null;
 
 Decoder::Decoder(World& world, std::string_view _id) :
   IdObject(world, _id),
+  m_functionLatchTimer{EventLoop::ioContext},
+  m_currentLatchedFunction(NO_FUNCTION),
   name{this, "name", "", PropertyFlags::ReadWrite | PropertyFlags::Store},
   interface{this, "interface", nullptr, PropertyFlags::ReadWrite | PropertyFlags::Store,
     [this](const std::shared_ptr<DecoderController>& value)
@@ -268,7 +271,7 @@ void Decoder::setFunctionValue(uint32_t number, bool value)
 {
   const auto& f = getFunction(number);
   if(f && getFunctionValue(f) != value)
-    f->value.setValueInternal(value);
+    f->value.setValue(value);
 }
 
 bool Decoder::acquire(Throttle& driver, bool steal)
@@ -441,4 +444,79 @@ void Decoder::changed(DecoderChangeFlags changes, uint32_t functionNumber)
   if(interface)
     interface->decoderChanged(*this, changes, functionNumber);
   decoderChanged(*this, changes, functionNumber);
+
+  if(has(changes, DecoderChangeFlags::FunctionValue))
+  {
+    const auto& f = getFunction(functionNumber);
+    if(m_currentLatchedFunction == functionNumber && (!f || f->value == false))
+    {
+      // Stop scheduled unlatch for current function, check other functions
+      rescheduleLatchedFunctionTimer();
+    }
+    else if(f && f->value == true && f->timeoutSeconds.value() > 0)
+    {
+      if(m_currentLatchedFunction == NO_FUNCTION)
+      {
+        // Schedule new timer
+        rescheduleLatchedFunctionTimer();
+      }
+      else if(m_currentLatchedFunction != functionNumber && f->getScheduledTimeout() < m_functionLatchTimer.expiry())
+      {
+        // If we timeout before current scheduled timer to this function
+        rescheduleLatchedFunctionTimer();
+      }
+    }
+  }
+}
+
+void Decoder::rescheduleLatchedFunctionTimer()
+{
+  if(m_currentLatchedFunction != NO_FUNCTION)
+  {
+    // Cancel current active timer before rescheduling
+    m_currentLatchedFunction = NO_FUNCTION;
+    m_functionLatchTimer.cancel();
+  }
+
+  std::chrono::steady_clock::time_point m_firstTimeout;
+
+  for(const auto& f : *functions)
+  {
+    if(f->timeoutSeconds.value() > 0 && f->value == true)
+    {
+      const auto scheduledTimout = f->getScheduledTimeout();
+      if(m_currentLatchedFunction == NO_FUNCTION || scheduledTimout < m_firstTimeout)
+      {
+        // We are the first function to timeout
+        m_firstTimeout = scheduledTimout;
+        m_currentLatchedFunction = f->number;
+      }
+    }
+  }
+
+  // Only start timer if needed
+  if(m_currentLatchedFunction != NO_FUNCTION)
+  {
+    m_functionLatchTimer.expires_at(m_firstTimeout);
+    m_functionLatchTimer.async_wait(std::bind(&Decoder::onLatchFunctionTimeout, this, std::placeholders::_1));
+  }
+}
+
+void Decoder::onLatchFunctionTimeout(const boost::system::error_code& ec)
+{
+  if(ec)
+    return; // Do nothing if we get cancelled
+
+  assert(m_currentLatchedFunction != NO_FUNCTION);
+
+  uint32_t currentLatchedFunction = m_currentLatchedFunction;
+
+  // Reset current latched function to allow rescheduling
+  m_currentLatchedFunction = NO_FUNCTION;
+
+  // We have not been cancelled, turn off latched function
+  const auto& f = getFunction(currentLatchedFunction);
+  f->value.setValue(false);
+
+  rescheduleLatchedFunctionTimer();
 }
--- a/server/src/hardware/decoder/decoderfunction.cpp
+++ b/server/src/hardware/decoder/decoderfunction.cpp
@@ -46,10 +46,15 @@ DecoderFunction::DecoderFunction(Decoder& decoder, uint8_t _number) :
     }},
   function{this, "function", DecoderFunctionFunction::Generic, PropertyFlags::ReadWrite | PropertyFlags::Store},
   value{this, "value", false, PropertyFlags::ReadWrite | PropertyFlags::StoreState,
-    [this](bool /*newValue*/)
+    [this](bool newValue)
     {
+      if(timeoutSeconds.value() > 0 && newValue)
+        m_scheduledTimeout = std::chrono::steady_clock::now() + std::chrono::seconds(timeoutSeconds.value());
+      else
+        m_scheduledTimeout = {};
       m_decoder.changed(DecoderChangeFlags::FunctionValue, number);
-    }}
+    }},
+  timeoutSeconds{this, "timeout_seconds", 0, PropertyFlags::ReadWrite | PropertyFlags::Store}
 {
   const bool editable = contains(decoder.world().state.value(), WorldState::Edit);
 
@@ -67,6 +72,9 @@ DecoderFunction::DecoderFunction(Decoder& decoder, uint8_t _number) :
   Attributes::addEnabled(value, true);
   Attributes::addObjectEditor(value, false);
   m_interfaceItems.add(value);
+  Attributes::addEnabled(timeoutSeconds, false);
+  Attributes::addMinMax(timeoutSeconds, 0, 60);
+  m_interfaceItems.add(timeoutSeconds);
 }
 
 std::string DecoderFunction::getObjectId() const
@@ -90,6 +98,9 @@ void DecoderFunction::worldEvent(WorldState state, WorldEvent event)
   Attributes::setEnabled(name, editable);
   Attributes::setEnabled(type, editable);
   Attributes::setEnabled(function, editable);
+
+  bool momentaryOrHold = (type == DecoderFunctionType::Momentary || type == DecoderFunctionType::Hold);
+  Attributes::setEnabled(timeoutSeconds, editable && momentaryOrHold);
 }
 
 void DecoderFunction::typeChanged()
@@ -110,4 +121,11 @@ void DecoderFunction::typeChanged()
       break;
   }
   Attributes::setEnabled(value, !isAlwaysOffOrOn(type));
+
+  const bool editable = contains(m_decoder.world().state.value(), WorldState::Edit);
+  bool momentaryOrHold = (type == DecoderFunctionType::Momentary || type == DecoderFunctionType::Hold);
+  Attributes::setEnabled(timeoutSeconds, editable && momentaryOrHold);
+
+  if(!momentaryOrHold)
+    timeoutSeconds.setValueInternal(0);
 }
--- a/server/src/hardware/decoder/decoderfunction.hpp
+++ b/server/src/hardware/decoder/decoderfunction.hpp
@@ -23,6 +23,7 @@
 #ifndef TRAINTASTIC_SERVER_HARDWARE_DECODER_DECODERFUNCTION_HPP
 #define TRAINTASTIC_SERVER_HARDWARE_DECODER_DECODERFUNCTION_HPP
 
+#include <chrono>
 #include "../../core/object.hpp"
 #include "../../core/property.hpp"
 #include "../../enum/decoderfunctiontype.hpp"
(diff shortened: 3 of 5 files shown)


server/src/hardware/decoder/decoderfunction.cpp
```diff
@@ -46,10 +46,15 @@ DecoderFunction::DecoderFunction(Decoder& decoder, uint8_t _number) :
     }},
   function{this, "function", DecoderFunctionFunction::Generic, PropertyFlags::ReadWrite | PropertyFlags::Store},
   value{this, "value", false, PropertyFlags::ReadWrite | PropertyFlags::StoreState,
-    [this](bool /*newValue*/)
+    [this](bool newValue)
     {
+      if(timeoutSeconds.value() > 0 && newValue)
```
reinder, line 51: Check here for `type == Momentary || Hold`? For other functions types the timeout value should not have effect.

server/src/hardware/decoder/decoderfunction.cpp
```diff
@@ -46,10 +46,15 @@ DecoderFunction::DecoderFunction(Decoder& decoder, uint8_t _number) :
...
+        m_scheduledTimeout = std::chrono::steady_clock::now() + std::chrono::seconds(timeoutSeconds.value());
+      else
+        m_scheduledTimeout = {};
       m_decoder.changed(DecoderChangeFlags::FunctionValue, number);
-    }}
+    }},
+  timeoutSeconds{this, "timeout_seconds", 0, PropertyFlags::ReadWrite | PropertyFlags::Store}
```
reinder, line 57: Should we change it to `ms` instead of `s`, e.g. 100ms steps? Maybe 1s steps is to big.

server/src/hardware/decoder/decoderfunction.cpp
```diff
@@ -90,6 +98,9 @@ void DecoderFunction::worldEvent(WorldState state, WorldEvent event)
   Attributes::setEnabled(name, editable);
   Attributes::setEnabled(type, editable);
   Attributes::setEnabled(function, editable);
+
+  bool momentaryOrHold = (type == DecoderFunctionType::Momentary || type == DecoderFunctionType::Hold);
+  Attributes::setEnabled(timeoutSeconds, editable && momentaryOrHold);
```
reinder, line 103: I prefer to hide properties that don't have effect due to other properties, disabling it might suggest it has effect but can't be changed at the moment. You can use the visible attribute for that.

server/src/hardware/decoder/decoderfunction.hpp
```diff
@@ -52,12 +55,22 @@ class DecoderFunction : public Object
     Property<DecoderFunctionFunction> function;
     Property<bool> value;
 
+    //! \brief Timeout in seconds to unlatch (turn off) active momentary functions
+    //! If set to zero, functions will not be automatically turned off
+    //! Active only for \ref DecoderFunctionType::Momentary and \ref DecoderFunctionType::Hold
+    Property<int> timeoutSeconds;
```
reinder, line 61: Traintastic preference is to use (u)intX_t types, in this case it can be unsigned as a negative value doesn't make sense.

reinder (commented):
Nice work, I like the single timer solution :)

reinder:
About the Z21 kernel change, what issue do you run into?

If `setValue` is called with the same value the property already is, it won't fire a change event, and it will not start the timer stuff (in case of a `true` value).

If communication with the has a big delay we can run into issues....

server/src/hardware/decoder/decoderfunction.cpp
gfgit: Timeout is reset to zero for all other function types but maybe it's more readable to check type instead

server/src/hardware/decoder/decoderfunction.cpp
gfgit: How do we achieve 100ms steps? Because 1ms step is too narrow. I've also noticed there is up to half second latency on turning off function so it would not respect milliseconds

gfgit:
> About the Z21 kernel change, what issue do you run into?
> 
> If `setValue` is called with the same value the property already is, it won't fire a change event, and it will not start the timer stuff (in case of a `true` value).
> 
> If communication with the has a big delay we can run into issues....

The problem is when external change happens.

1. Let's say you have F1 off.
2. Z21 changes F1 to on.
3. Decoder is updated from `ClientKernel` and calls `setValue()`
4. This changes from Off to On so it triggers `onChanged()` callback
5. The callback calls `Decoder::changed()`
6. This calls interface's `decoderChanged()`
7. We send back to Z21 a message saying to turn on F1 which is useless because it already knows...

The boolean guard `m_isUpdatingDecoderFromKernel` prevents sending back to Z21 a change that was triggered by Z21 itself. This reduces traffic and also in case of latency or rapid turn on/off sequence it behaves correctly.
It should be done also for other kernels.
The concept here is that Z21 maintains internal state and it's the Master. So we should only acknowledge changes coming from Z21 and not mirror them back to it.
If a change is made from Traintastic instead, the `m_isUpdatingDecoderFromKernel` will be false so the change gets propagated to Z21 correctly and Z21 will reply to us with the same change.
Since the change was originated by us, we do not care about the reply and that's why I made PR #73 so that we ignore replies to our own changes since we already know the new value.

server/src/hardware/decoder/decoderfunction.cpp
gfgit: Right. Done

server/src/hardware/decoder/decoderfunction.hpp
gfgit: Done. It's now `uint16_t` in preparation for milliseconds.

server/src/hardware/decoder/decoderfunction.cpp
gfgit: I've left the enabled code so it prevents setting values to it.

server/src/hardware/decoder/decoderfunction.cpp
reinder: Yeah, I prefer to check type, the we don't need to alter the timeout, if someone accidentally changes the type it keeps the timeout value when set back.

server/src/hardware/decoder/decoderfunction.cpp
reinder: rounding to 100ms can be done using the Properties `onSet` callback and we should add a `Step` attribute, we can set it to 100, the the client can take that into account.

server/src/hardware/decoder/decoderfunction.cpp
```diff
@@ -125,6 +127,7 @@ void DecoderFunction::typeChanged()
   const bool editable = contains(m_decoder.world().state.value(), WorldState::Edit);
   bool momentaryOrHold = (type == DecoderFunctionType::Momentary || type == DecoderFunctionType::Hold);
   Attributes::setEnabled(timeoutSeconds, editable && momentaryOrHold);
+  Attributes::setVisible(timeoutSeconds, editable && momentaryOrHold);
```
reinder: Enabled should only be affected by `editable`
Visible should only be affected by `momentaryOrHold`

Then it behaves the same like other properties. (I should document that some where...)

server/src/hardware/decoder/decoder.cpp
```diff
@@ -482,7 +482,7 @@ void Decoder::rescheduleLatchedFunctionTimer()
 
   for(const auto& f : *functions)
   {
-    if(f->timeoutSeconds.value() > 0 && f->value == true)
+    if(f->hasTimeout() > 0 && f->value == true)
```
reinder, line 485: `hasTimeout()` is a nice solution, but you should remove the `> 0`, that is confusing :P

server/src/hardware/decoder/decoderfunction.hpp
```diff
@@ -74,6 +74,8 @@ class DecoderFunction : public Object
     {
       return m_scheduledTimeout;
     }
+
+    inline bool hasTimeout() const { return timeoutSeconds.value() > 0; }
```
reinder, line 78: Shouldn't the type check be added here too?

server/src/hardware/decoder/decoderfunction.cpp
gfgit: Done.

server/src/hardware/decoder/decoderfunction.cpp
gfgit: Done.

server/src/hardware/decoder/decoderfunction.cpp
gfgit: Done.

reinder:
> > About the Z21 kernel change, what issue do you run into?
> > If `setValue` is called with the same value the property already is, it won't fire a change event, and it will not start the timer stuff (in case of a `true` value).
> > If communication with the has a big delay we can run into issues....
> 
> The problem is when external change happens.
> 
>     1. Let's say you have F1 off.
> 
>     2. Z21 changes F1 to on.
> 
>     3. Decoder is updated from `ClientKernel` and calls `setValue()`
> 
>     4. This changes from Off to On so it triggers `onChanged()` callback
> 
>     5. The callback calls `Decoder::changed()`
> 
>     6. This calls interface's `decoderChanged()`
> 
>     7. We send back to Z21 a message saying to turn on F1 which is useless because it already knows...
> 
> 
> The boolean guard `m_isUpdatingDecoderFromKernel` prevents sending back to Z21 a change that was triggered by Z21 itself. This reduces traffic and also in case of latency or rapid turn on/off sequence it behaves correctly. It should be done also for other kernels. The concept here is that Z21 maintains internal state and it's the Master. So we should only acknowledge changes coming from Z21 and not mirror them back to it. If a change is made from Traintastic instead, the `m_isUpdatingDecoderFromKernel` will be false so the change gets propagated to Z21 correctly and Z21 will reply to us with the same change. Since the change was originated by us, we do not care about the reply and that's why I made PR #73 so that we ignore replies to our own changes since we already know the new value.

Ok, I understand.

I think we should fix this in the `DecoderFunction`, when the value gets updated by all the kernels it should do the timeout magic and use `setValueInternal` like before, so it doesn't trigger the `onChanged`, else we need this kind of logic in every kernel which is error prone.

gfgit:
> I think we should fix this in the `DecoderFunction`, when the value gets updated by all the kernels it should do the timeout magic and use `setValueInternal` like before, so it doesn't trigger the `onChanged`, else we need this kind of logic in every kernel which is error prone.

In my opinion things should stay this way.
This boolean guard logic was already needed for Decoder throttle property, I've just extended it to cover functions.
Otherwise for every change we need to know where is coming from instead of having a generic onChanged callback in which you can throw everything.
The complexity I think would be greater this way than a simple boolean guard, but we could have a draft code to compare better.
An example of this is Output. When operated from Kernel (external change) it calls `updateOutputValue()` which uses `setValueInternal()` and manually fires events.
When Traintastic changes an output, it calls `setValue()` which goes to interface which is responsible to call `updateOutputValue()`.
In case of Z21 this is async, meaning `setValue()` will generate a message to Z21 and when we receive a reply we do `updateOutputValue()` part.
Position of the Output is updated instantaneously but events are fired only if we get a reply.

For Decoder properties we do not have events to fire and we consider the change to happen just when we trigger it without waiting for any reply.

reinder:
I'm still not a fan of the boolean guard, besides that is must be implemented in every protocol kernel there may be other side effects.

The guard currently block *all* decoder changes, so if a function value change somehow triggers another decoder change it gets lost/blocked.

The output systems work indeed a bit different, that's mainly because I've rewritten it (because it wasn't working very well), I think the Decoder / Input part needs a similar rewrite. So I think we should apply that here too.

Make `value` read only and add a `setValue` method and an `updateValue` function, just like the Outputs.
`value` reflecting the current hardware value and using `setValue` to request the change, once the hardware processed it it will update the `value`.

The same applies to the `Decoder` for speed/direction etc. But that should be a different issue/PR :)

I fact we should use this *"pattern"* for all such hardware values.

Please let me know what your thoughts are on this :)

#BuildBetterSoftwareTogether

gfgit:
> I fact we should use this _"pattern"_ for all such hardware values.

Then we could open a draft PR and see where this pattern lead us to. Maybe it's less complicated than I've imagined

reinder:
I will look into it :) might take a few days, need to find some time for it.

reinder:
Just did a quick scan, the `value` / `setValue` pattern doesn't fit currently, as the `DecoderController` currently just gets notified about the change, the value is read from the `value` property, but that is what we should get rid of...so it requires some refactor to integrate that.

For this PR I think adding a `updateValue()` to `DecoderFunction` would be wise, it can do the timer stuff if required and call `value.setValueInternal`, the the guard stuff in the Z21 kernel can be reverted to only decoder stuff. All protocol kernels the should call `updateValue` instead of `value.setValueInteral`. I expect that will work, but we need to test it off course :)

gfgit:
@reinder Hi, I've rebased on current master and tried your `updateValue()` suggestion.
I like the principle but the problem is still relevant.

Interfaces call `Decoder::setFunctionValue()` which internally now calls `DecoderFunction::updateValue()` instead of property `setValue()`.
This excludes the `OnChanged()` callback but part of it is fine.
Inside on `OnChanged()` it calls `Decoder::changed()` which then calls `interface->decoderChanged()`.

The point is not calling `interface->decoderChanged()` if the change was originally triggered by interface kernel.
But `Decoder::changed()` also does other things so by not calling it we miss emitting `Decoder::decoderChanged()` and function latch timer schedule (which I had to put also in other places to ensure it gets triggered).

So maybe the best thing is adding an argument to `Decoder::changed()` to skip updating interface and let it do all other things the same way.

reinder:
> Just did a quick scan, the value / setValue pattern doesn't fit currently, as the DecoderController currently just gets notified about the change, the value is read from the value property, but that is what we should get rid of...so it requires some refactor to integrate that.

I still think this is the way to go (for all hardware/control properties), a read only `value` property and a `setValue` method, in case of the decoder function an extra function with a time param could be added too. Using a method for changing something makes it also possible to block it under some conditions and for example report the reason. (very useful for scripting too.) For the webthrottle (#178) I made modifications to the `Throttle` to implement a separate set path it e.g. block changing direction if the train is moving.

If this works for now I'll merge it, we can do the refactor later. I think we should critically review the current Decoder interfacing, it was one of the first things I built.

gfgit:
I've quickly read #178
Well, we are not that far from it.
Now interfaces use `updateDirection/Throttle/EStop/Function()` methods.

And rest of the code should use direction.setValue().
But it's not explicitly used anywhere so it's just the client setting exposed properties on the object.

I can make properties read only. Then I think new setValue() function could carry the boolean parameter and pass it down to `Decoder::changed()`.
So we have just one method for both interfaces and client.

gfgit:
If I change `DecoderFunction::value` to be ReadOnly, and add a `Method<bool(bool)> updateValue`, then I should change client `ThrottleFunctionButton` code to call method instead of setting property value, right?

Also I'm concerned about `Decoder::getFunctionValue()`.
Let's say you have set "No Smoke" world mode, there is no explicit `setValue(false)` for smoke functions and then their value is ignored and function returns false.
But now what if user externally activates smoke function via hardware throttle? Should we reset it to false automatically?
When "No Smoke" mode gets disabled, should we remember previous state of the function and reactivate it if it was disabled by us?

Anyway returning a fake value inside `Decoder::getFunctionValue()` seems confusing to me.

reinder:
> I've quickly read [#178](https://github.com/traintastic/traintastic/issues/178)
> Well, we are not that far from it.
> Now interfaces use `updateDirection/Throttle/EStop/Function()` methods.

The web throttle uses an explicit aquire/steal/release, similar to WiThrottle. For the traintastic-client trottle I'd like to use the same, so there is only one Throttle in control. For furure automatic control it then makes sense that a System/AutoPilot throttle is assigned to the train. If some throttle than steals it, automatic mode can be disabled automatically.

Currently there is also an decoder throttle, I'd like to get rid of that, I created it because Trains weren't there yet.

> And rest of the code should use `direction.setValue()`.

If a property is ReadOnly, `setValue` can't be used, you then must use `setValueInternal`.

> If I change `DecoderFunction::value` to be ReadOnly, and add a `Method<bool(bool)> updateValue`, then I should change client `ThrottleFunctionButton` code to call method instead of setting property value, right?

Yes that's right, I prefer to name the method `setValue`, thats also used for turnouts/signals etc.

> Also I'm concerned about Decoder::getFunctionValue().
> Let's say you have set "No Smoke" world mode, there is no explicit setValue(false) for smoke functions and then their value is ignored and function returns false.
> But now what if user externally activates smoke function via hardware throttle? Should we reset it to false automatically?
> When "No Smoke" mode gets disabled, should we remember previous state of the function and reactivate it if it was disabled by us?

The value is the user value, if the no smoke function changes it to false, then what if the user turns it on, should it enable? When smoke is allowed should it turn back on?

> Anyway returning a fake value inside Decoder::getFunctionValue() seems confusing to me.

This function should be used by the interface/kernel to get the value that must be sent to the track. Maybe we should rename it then. I don't want to handle the mute/no smoke stuff in the interface because then we need to implement that many times. In the future I'd like to also have blocks/zones which can be marked mute/no smoke (e.g. the shadow station).
So it basically returns the function value based on the user value and optional limitations dictated by the world/block/zone etc.

Maybe we should add multiple read only function `value` properties, one for the user value and one for the "track" value. The information can the also be shown in the throttle.

For train speed we problably have something similar, actual speed, target speed, speed limit etc.

PR #163, opened 2024-08-16
Sync speed on all Train vehicles

gfgit:
This revives PR #54 which got closed by mistake.

Basically it allows to have per Train-Vehicle properties, like which vehicle are powered on inside the Train and which are not.
Of course all model locomotives are powered on and moving their motors but the effect of one locomotive pulling the other can be simulated by turning on and off engine sound if present and lights.

It still lacks the speed matching table for which I need a little help to find a good implementation.
Basically it should be a vector property per each Decoder, inherited by PowerVehicle.
Then Train should mix all of these properties and compute a new table specific to current consist formation.


reinder:
When I load a file created with a build that doesn't have this change, the train vehicle list is empty when the loaded world contains trains with assigned vehicles.

Can you add some code for that so it loads properly? (once saved it can not be loaded in an old build, but that's not an issue)

As an example, there are some loading fixes for old files in `WorldLoader::createObject` or `SensorRailTile::load`, I mark them with `//! \todo Remove in v0.4` (Traintastic only supports *upgrades* one minor version back until v1.0)
Once v0.3 is released I can remove them easily.

gfgit:
> Can you add some code for that so it loads properly? (once saved it can not be loaded in an old build, but that's not an issue)

Done.
Since we are at it, should we make `RailVehicle::activeTrain` a `NoStore` like `RailVehicle::trains` property and instead calculate it inside `Train::load()` ?
I think this makes it more robust against file corruption. Let me know!

gfgit:
Rebased on latest master to resolve conflicts

reinder:
> Since we are at it, should we make `RailVehicle::activeTrain` a `NoStore` like `RailVehicle::trains` property and instead calculate it inside `Train::load()` ? I think this makes it more robust against file corruption. Let me know!

Yes please, good point.

If we have a corrupt file and a vehicle is activated twice due to being in two active trains, I think the first one should win and the other train must me deactivated and log a warning/error.

If that second train is in one or more blocks, the train must be removed from those blocks. A train must be active to be in a block. (I think a train can be active without any block, because else you can't test drive a train easily.)

gfgit:
> If that second train is in one or more blocks, the train must be removed from those blocks. A train must be active to be in a block. (I think a train can be active without any block, because else you can't test drive a train easily.)

How does this relate to shunting. Like a train stops at a station platform and gets cut into 2 parts which become 2 trains on same block.

reinder:
> How does this relate to shunting. Like a train stops at a station platform and gets cut into 2 parts which become 2 trains on same block.

The "uncut" train is deactivated and two new (temporary*) trains are created (or the exact matching one is activated) , then those two are active.

* temporary trains are auto deleted on deactivation.

(just my current vision on train splitting/shunting)

reinder:
@gfgit Nice work!

As the file is text we can sent a `std::string`, for the read from / save to file some logic on the client side is required. See the import/export world code, that is also a bit special. Most of Traintastic's UI is automatically built, but thats a nice to have to make it easier, if custom code is needed to improve the UX the we should do that.

My suggestion would be to add two methods, something like:
`Method<void(SpeedTableFormat,std::string)> import`
`Method<std::string(SpeedTableFormat)> export`

Then we can easily support other formats too, e.g. CSV

For the JSON format I prefer a sparse table, this is much more flexible.

For the missing 126 we have to options, limit loco max speed so it isn't used (or interplate based on the previous two values.)
For the lowest value, something similar should be available, e.g. not all loco's drive properly at speedstep 1, so a minimum may be viable there too.

gfgit:
Thanks for the suggestions.
Also what about Decoders not having 126 steps?
The algorithm is quite generic but still hardcodes to 126 steps.
Maybe we can keep it like that until it works properly and then expand it.

gfgit:
Another problem is that Train properties `speed`, `speedMax` and more importantly `throttleSpeed` become discrete instead of continuous.
This means Client throttle widget will set a value to `throttleSpeed` and that value needs to be rounded.
It may happen to be rounded back to previous value making it impossible to change value.

This also happens with decoder speed steps coming from external throttle.
In this case I've solved the problem by applying a delay timer. In this period of time user can keep moving the throttle, rounding will happen on timeout.
There is also smart rounding which will round up/down based on speed trend which makes the delay only need on special circumstances.

reinder:
> Thanks for the suggestions. Also what about Decoders not having 126 steps? The algorithm is quite generic but still hardcodes to 126 steps. Maybe we can keep it like that until it works properly and then expand it.

I think it is best to import the table as is. If it has 28 steps convert speed to a 0...126 step value (no rounding) than convert that to 0..28 and round it. (Log a warning on import and world load would be nice.)

> Another problem is that Train properties `speed`, `speedMax` and more importantly `throttleSpeed` become discrete instead of continuous. This means Client throttle widget will set a value to `throttleSpeed` and that value needs to be rounded. It may happen to be rounded back to previous value making it impossible to change value.
>
> This also happens with decoder speed steps coming from external throttle. In this case I've solved the problem by applying a delay timer. In this period of time user can keep moving the throttle, rounding will happen on timeout. There is also smart rounding which will round up/down based on speed trend which makes the delay only need on special circumstances.

What if we just start with a blank sheet of paper and write down how we think it should work.
- How to handle hardware throttles. (2 flavors: e.g. WiThrottle or direct affecting the command station)
- How to handle software throttles. (the dialog in Traintastic client)
- How to handle automatic control. (for future automatic driving)
- Which properties should a train have: `speedMin`?, `speedMax`, `actualSpeed`, `targetSpeed`/`trottleSpeed`?
- Instead of setting the target speed, a seperate power and brake setting could also be nice.

For WiThrottle I implemented a seperate Throttle object, because it supports explicit control/stealing.
What if we implement such a system, that there is only one (or no) throttle that is in control of the train,
then we have a kind of ownership which makes it easier to handle conflicts. For hardware throttles that don't support stealing we can implement *auto steal* (enable/disable) or use a Function key for stealing control.

Just a lot of ideas, feel free to share yours and give feedback on mine :)

#BuildBetterSoftwareTogether

server/src/train/train.hpp
```diff
@@ -79,10 +79,10 @@ class Train : public IdObject
 
     //! \todo add realistic acceleration
     //! \note m/s^2 in physical model scale
-    double m_accelerationRate = 1.5; // m/s^2
+    double m_accelerationRate = 1.5 / 87.0; // m/s^2
 
     //! \todo add realistic braking
-    double m_brakingRate = -1.0; // m/s^2
+    double m_brakingRate = -1.0 / 87.0; // m/s^2
```
reinder: Is this 87 as in H0 1:87? If so `world.scaleRatio` should be used.

gfgit:
Yes. It was just a quick fix.
ScaleRatio is already used elsewhere.
But if user sets it to "custom" it becomes zero and so would crash on divisions...

gfgit:
Yes we should write it down.
For now I think external throttle is quite well handled.
Stealing is a different subject.
I mean, first I want it to be compatible with all kinds of throttle (there are 2/3 main types I think, like rotary knob with fixed andgle, step by step, infinite rotating knob...).
After we can restrict driving through ownership and stealing.

For WiThrottle we can short circuit the logic and make it directly control the whole Train via `setThrottleSpeed()` which should be renamed to target speed.
For external throttle this is not possible of course, so my logic introduces rounding, a bit of delay for special cases and in future maybe some hysteresis to improve rotary knobs.

For Traintastic throttle widget it could be same thing as WiThrottle but since you can already use Train throttle for this, we could leave Decoder throttle mimic an external throttle and so go through the same rounding logic.
This is handy to debug in simulation mode and other scenarios.

As for supporting other speed step ranges I need to think more.
Maybe we could store tables always as sparse and interpolate at world loading.
Then we could just use it. There is no real limit of having same step range on all locomotives, I just need to tweak the loop range for each.

Speed min is good to have.

Speed max could be divided.
There is maximum matched speed which is absolute maximum.
Then there is maximum speed of non powered cars, like you couple a particular tank wagon which imposes a speed limitation.

Then there is line (per block?) speed limit.

But there is also acceleration to take in account here, as in "max speed to be able to stop inside a block length".
In fact in Italy, trains running at > 150 km/h know the aspect of next 2 signals.
Therefore only by knowing next 2 signals are green you can go at 200 km/h, and you start braking at green signal because you already know next will be yellow and the one after is red.
So you can brake in 2 blocks length instead of one and this allows greater speeds.
But on lines not equipped to inform train driver of the aspect of next 2 sognals, speed is limited to 150. Numbers could be a bit different but you got the idea.

We could calculate them all, the take the mininmun, store it and use it for limiting in SemiAutomatic mode.
Manual mode could be limited only by speed matching max speed.
Automatic mode will also decelerate and stop at red?

I like the idea of stealing through function keys. Also automatic/manual mode could be switched by function keys.

How would you use power and brake instead of throttle?

reinder:
Sorry for the delay, I was on a work trip.

> Yes. It was just a quick fix. ScaleRatio is already used elsewhere. But if user sets it to "custom" it becomes zero and so would crash on divisions...

If it can becomes zero somewhere, we should fix that, the range is `Attributes::addMinMax(scaleRatio, 1., 1000.);`

For this PR it is best I think to fix the command station throttle changes, we can leave WiThrottle as it is for now.
Then it should be possible to drive on the layout with the command station throttle and Traintastic should be able to track that train across the layout. That is really cool! :smile:

For min/max speed and manual/semi auto mode we're on the same page :)

In auto mode, Traintastic controls the targetSpeed, the user can only take control back by setting the mode to manual or semi auto. So if a hardware throttle modifies the speed Traintastic will set it back. (In case of a WiThrottle steal, then mode can changed to manual or semi auto, that can be a config option.)
So when in auto mode, Traintastic attaches a "virtual driver" throttle (we need to build that) to the train.

I still really like the idea of having one throttle/driver that is in control of the train. There are three kinds:
- Stand alone (Traintatic is always between the throttle and the command station):
  - Traintastic Client throttle dialog
  - WiThrottle
  - WLANmaus (if connected to Traintastic)
  - Traintastic Android/iOS/web app
- Command station, throttles that directly control locomotives because the connect to the command station directly (these are the hardest ones)
- Auto pilot, for Automatic mode

server/src/vehicle/rail/poweredrailvehicle.cpp
```diff
@@ -36,6 +36,8 @@ PoweredRailVehicle::PoweredRailVehicle(World& world, std::string_view id_)
   : RailVehicle(world, id_)
   , power{*this, "power", 0, PowerUnit::KiloWatt, PropertyFlags::ReadWrite | PropertyFlags::Store}
   , speedCurve{this, "speed_curve", nullptr, PropertyFlags::ReadOnly | PropertyFlags::SubObject | PropertyFlags::Store | PropertyFlags::ScriptReadOnly}
+  , maxAccelerationRate{this, "max_acceleration_rate", 0, PropertyFlags::ReadWrite | PropertyFlags::ScriptReadOnly | PropertyFlags::Store}
+  , maxBrakingRate{this, "max_braking_rate", 0, PropertyFlags::ReadWrite | PropertyFlags::ScriptReadOnly | PropertyFlags::Store}
```
reinder, line 40: default should not be zero I think

server/src/vehicle/rail/poweredrailvehicle.cpp
```diff
@@ -44,9 +46,19 @@ PoweredRailVehicle::PoweredRailVehicle(World& world, std::string_view id_)
   m_interfaceItems.add(power);
 
   speedCurve.setValueInternal(std::make_shared<VehicleSpeedCurve>(*this, speedCurve.name()));
-  Attributes::addEnabled(speedCurve, true);
+  Attributes::addEnabled(speedCurve, editable);
   m_interfaceItems.add(speedCurve);
 
+  Attributes::addDisplayName(maxAccelerationRate, DisplayName::Vehicle::Rail::maxAccelerationRate);
+  Attributes::addEnabled(maxAccelerationRate, editable);
+  Attributes::addMinMax(maxAccelerationRate, 0.0, 10.0); // m/s^2
```
reinder, line 54: minimum should be > 0

server/src/vehicle/rail/poweredrailvehicle.cpp
```diff
@@ -44,9 +46,19 @@ PoweredRailVehicle::PoweredRailVehicle(World& world, std::string_view id_)
...
+  m_interfaceItems.add(maxAccelerationRate);
+
+  Attributes::addDisplayName(maxBrakingRate, DisplayName::Vehicle::Rail::maxBrakingRate);
+  Attributes::addEnabled(maxBrakingRate, editable);
+  Attributes::addMinMax(maxBrakingRate, -10.0, 0.0); // m/s^2
```
reinder, line 59: maximum should be < 0

gfgit:
I've put zero for cases where you don't know physical train acceleration and so it does not partucipate to acceleratio of Train.
Train max acceleration will be min of all vehicles except zeros.
It is weird to set it zero, yes, but then what do we do for trsins which we don't know acceleration?

gfgit:
My idea is max accel/braking is what physically can do the locomotive. For example if decoder has some inertia settings.
A little bit of decoder inertia is good to smooth Train acceleration when driven from external throttle as it gives time to Traintastic to ovverride the target speed set by user with current speed slowly increasing.

So even setting ridiculous values in power field we should no exceed max acceleration.
It's like a safe max. Otherwise locomotives in the consist might get out of sync during acceleration phase.

I've estimated real passenger train acceleration to somewhat like 0.5 m/s^2.
It's just a real quick calculation but it's already way lower than what I've set by default.
Which again is lower than my decoder inertia (it could accelerate way faster).

BUT... 0.5 m/s^2 is not really usable value. You would need a very big layout just to reach top speed, and what about braking?
Realism is good but here I think we need to compromise while letting users witj big layout to set more realistic values.
So power/weight calculation would lead to realistic values but not really usable if we get the maths correct.

gfgit:
Another option is to calculate realistic acceleration/braking rate based on Power/Weight and other parameters.
Then we have a global acceleration factor which multiplies the result to shorten distances for small layouts.

I've found an interesting website for the math part:
https://www.engineeringtoolbox.com/car-acceleration-d_1309.html
https://www.engineeringtoolbox.com/lift-drag-fluid-flow-d_1657.html
https://www.engineeringtoolbox.com/friction-coefficients-d_778.html
https://www.engineeringtoolbox.com/moment-rotating-shaft-d_952.html

Anyway I would keep this calculations for later PRs because this one is already complex.

reinder:
> My idea is max accel/braking is what physically can do the locomotive. For example if decoder has some inertia settings. A little bit of decoder inertia is good to smooth Train acceleration when driven from external throttle as it gives time to Traintastic to ovverride the target speed set by user with current speed slowly increasing. So even setting ridiculous values in power field we should no exceed max acceleration. It's like a safe max. Otherwise locomotives in the consist might get out of sync during acceleration phase. I've estimated real passenger train acceleration to somewhat like 0.5 m/s^2. It's just a real quick calculation but it's already way lower than what I've set by default. Which again is lower than my decoder inertia (it could accelerate way faster). BUT... 0.5 m/s^2 is not really usable value. You would need a very big layout just to reach top speed, and what about braking? Realism is good but here I think we need to compromise while letting users witj big layout to set more realistic values. So power/weight calculation would lead to realistic values but not really usable if we get the maths correct.

OK, I got it, I agree :)

I've read about a multiplication factor in other software to, the have that for the same reason, so we should do that too :)

So for now we use these values, in the future we can add more realistic behavior and use these values as limits.

server/src/train/train.cpp
```diff
@@ -186,6 +188,14 @@ Train::Train(World& world, std::string_view _id) :
   Attributes::addObjectEditor(throttleSpeed, false);
   m_interfaceItems.add(throttleSpeed);
 
+  Attributes::addMinMax(accelerationRate, 0.01, 10.0);
+  Attributes::addUnit(accelerationRate, "m/s^2");
+  m_interfaceItems.add(accelerationRate);
+
+  Attributes::addMinMax(brakingRate, -10.0, 0.01);
```
reinder, line 195: max must be `-0.01` I think.

server/src/train/train.cpp
```diff
@@ -186,6 +188,14 @@ Train::Train(World& world, std::string_view _id) :
   Attributes::addObjectEditor(throttleSpeed, false);
   m_interfaceItems.add(throttleSpeed);
 
+  Attributes::addMinMax(accelerationRate, 0.01, 10.0);
+  Attributes::addUnit(accelerationRate, "m/s^2");
```
reinder, line 192: Can we use a superscript two here using UTF8? see: https://github.com/traintastic/traintastic/blob/master/server/src/utils/utf8.hpp

server/src/vehicle/rail/poweredrailvehicle.cpp
```diff
@@ -51,12 +51,12 @@ PoweredRailVehicle::PoweredRailVehicle(World& world, std::string_view id_)
...
   Attributes::addDisplayName(maxAccelerationRate, DisplayName::Vehicle::Rail::maxAccelerationRate);
   Attributes::addEnabled(maxAccelerationRate, editable);
-  Attributes::addMinMax(maxAccelerationRate, 0.0, 10.0); // m/s^2
+  Attributes::addMinMax(maxAccelerationRate, 0.01, 10.0); // m/s^2
   m_interfaceItems.add(maxAccelerationRate);
 
   Attributes::addDisplayName(maxBrakingRate, DisplayName::Vehicle::Rail::maxBrakingRate);
   Attributes::addEnabled(maxBrakingRate, editable);
-  Attributes::addMinMax(maxBrakingRate, -10.0, 0.0); // m/s^2
+  Attributes::addMinMax(maxBrakingRate, -10.0, 0.01); // m/s^2
```
reinder, line 59: max = `-0.01` ?

server/src/train/trainvehiclelistitem.cpp
```diff
@@ -58,9 +58,6 @@ TrainVehicleListItem::TrainVehicleListItem(const std::shared_ptr<RailVehicle> &v
 
 TrainVehicleListItem::~TrainVehicleListItem()
 {
-  assert(!vehicle);
-  assert(!m_vehiclePropertyChanged.connected());
-  assert(!m_vehicleDestroying.connected());
```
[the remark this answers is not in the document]
gfgit: Fixed in last commit!

gfgit:
@reinder Now that Vehicles store every Train they are part of in `trains` property...
We could move the removal of deleted Vehicles to Vehicle itself inside `destroying()`.
This avoids a connection from `TrainVehicleListItem` to `Vehicle::onDestroyed()`.
But maybe code would become more complex.

server/src/vehicle/rail/poweredrailvehicle.cpp
```diff
@@ -36,8 +36,8 @@ PoweredRailVehicle::PoweredRailVehicle(World& world, std::string_view id_)
   : RailVehicle(world, id_)
   , power{*this, "power", 0, PowerUnit::KiloWatt, PropertyFlags::ReadWrite | PropertyFlags::Store}
   , speedCurve{this, "speed_curve", nullptr, PropertyFlags::ReadOnly | PropertyFlags::SubObject | PropertyFlags::Store | PropertyFlags::ScriptReadOnly}
-  , maxAccelerationRate{this, "max_acceleration_rate", 0, PropertyFlags::ReadWrite | PropertyFlags::ScriptReadOnly | PropertyFlags::Store}
-  , maxBrakingRate{this, "max_braking_rate", 0, PropertyFlags::ReadWrite | PropertyFlags::ScriptReadOnly | PropertyFlags::Store}
+  , maxAccelerationRate{this, "max_acceleration_rate", 3.0, PropertyFlags::ReadWrite | PropertyFlags::ScriptReadOnly | PropertyFlags::Store}
+  , maxBrakingRate{this, "max_braking_rate", 2.0, PropertyFlags::ReadWrite | PropertyFlags::ScriptReadOnly | PropertyFlags::Store}
```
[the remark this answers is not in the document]
gfgit: Right, fixing it now!

gfgit:
What if we made braking rate positive and then add a minus in acceleration calculations?
I think it's counter intuitive to work with negative values because you reverse every
less than/greater than checks and reverse min/max meaning.
Also for the inexperienced user it may be easier to just work with positive numbers.
What do you think?

server/src/train/train.cpp
gfgit: Done this as welllllll.
see it's complicated working on negatives

reinder:
> What if we made braking rate positive and then add a minus in acceleration calculations?
> I think it's counter intuitive to work with negative values because you reverse every
> less than/greater than checks and reverse min/max meaning.
> Also for the inexperienced user it may be easier to just work with positive numbers.
> What do you think?

Yeah, that's much easier for the user and also less error prone in code.

reinder:
> @reinder Now that Vehicles store every Train they are part of in `trains` property...
> We could move the removal of deleted Vehicles to Vehicle itself inside `destroying()`.
> This avoids a connection from `TrainVehicleListItem` to `Vehicle::onDestroyed()`.
> But maybe code would become more complex.

I prefer that, then we don't need to connect to onDestroyed. I think it will make it easier to understand, it also simplifies loading code, as there is no need for connecting.

For trains and blocks it works similar if I remember correctly.

server/src/train/train.cpp
reinder: yeah, all positive like you suggested is much better :)

reinder:
@gfgit Can you add one (or more) speed table files here, just so I can try it :)

reinder:
Hi,

It is affected a bit, the `Decoder` now has a `vehicle` property which is automatically set when you assign a decoder to a `RailVehicle`. So that can be used now. Before it was possible to assign the same decoder to different rail vehicles.

A throttle is acquired on a train, not on a locomotive, so it doesn't matter how many loco's there are in a train.

I think we should add a setting how traintastic should handle manual interference of the train speed/direction. I expect that it will differ per user/system.

gfgit:
Also there is a problem with "immediate" throttle commands.
Since locomotives in a train can have different acceleration values, usually speed cannot be applied immediately.
Otherwise locomotives with higher acceleration rate will reach it before other locomotives in the same train.
We also go through every matched speed point between current speed and target speed, instead of just setting target speed.

Train class knows this and has a logic to accelerate at the lowest acceleration rates of train's locomotives.
Acceleration rate can (and should!) be further clamped down by user to get realistic values.
So a possible solution would be:
- set target speed follows train acceleration rate (calculated with lowest loco rate + user value)
- immediate set speed follows just lowest loco rate, ignoring possible user reduction to acceleration values.

Same goes for braking.

If you think about it, even without this PR, if user set acceleration rate on decoder CVs, and used "immediate" throttle commands, the locomotive still would go through acceleration instead of immediately reaching target speed.
So we could think acceleration split in 2 cases:
- Acceleration set into decoder / physical locomotive acceleration
- Acceleration emulated by traintastic logic

Immediate commands should ignore second one, but first type will always be present and in multiple traction scenario it must be taken into account to sync all locomotives

Issue #156, opened 2024-07-08
Weird UI Bug - Interfaces


DrEVILish:
There is also a weird UI bug
Where I add a new Interface and the UI Interfaces table doesn't update. If I change the name or ID, it then populates the table. When it's like this I can't add or edit or delete the item as I'm guessing it's "ID" is missing. I have to make sure I "edit" the ID before closing that Interface window.
Let me know if you want more debug on this:
<img width="804" alt="Screenshot 2024-07-08 at 13 50 20" src="https://github.com/traintastic/traintastic/assets/6287003/28eaafd8-a9b0-452d-a98f-581c44822d8d">

reinder:
That is indeed a strange issue (I'd expect it to be some kind of timing issue.):
- If you close the Interfaces dialog and open it again, is the row still empty then.
- Does it happen every time?
- Are you running a *Debug* or *Release* build?

DrEVILish:
If I close and reopen, it displays the correct information and can delete, edit and add the items again.

Happens every time repeatable.

Everything I'm "testing" with is built using the BUILDING.md instructions with a small modification for multicore building (faster). "Release" AFAIK, always `git pull` latest from here. 
Except for Qt6 instead of Qt5, could be a possible Qt versioning bug.

reinder:
Thanks for the info, I'll try with Qt6 tomorrow (if I have time), I'm on Ubuntu 22.04 with Qt5. Now time to get some sleep, its midnight here :) (I live in the Netherlands.)

reinder:
Just checked with Qt 6.2.4 on Ununtu 22.04, no issue.

Is it possible for you to do some debugging?

gfgit:
I've done some debugging on `TableModel` which should apply also here.
The problem is that `Region` object does not store negative values. Also it's represented with `uint32_t` on server side and with `int` on Client side so there is also a possible type mismatch.
Also there is no difference between a empty model (rowCount == 0) and a model with a single item in it. `Region`'s `rowMin` and `rowMax` are both equal to zero.
So when adding a new item, Client checks if it should request a region update but regions are equal so it skips update.
Row count is updated so a new row is shown in Client table view but empty because region update was skipped.

To avoid this we may use `rowMin` = 1 and `rowMax` = 0 to have an "invalid" region which represent empty model.

DrEVILish:
Unable to recreate via the previous method, Consider #165 as fixed

DrEVILish:
[...]

PR #165, opened 2024-08-16
Fix Table Model

gfgit:
This is my attempt to fix issue #156 
I'm not 100% sure it fixes adding new items to table model because from time to time it does not work.
@DrEVILish if you can please test this patch :)

### DIFF
--- a/client/src/network/connection.cpp
+++ b/client/src/network/connection.cpp
@@ -425,7 +425,7 @@ void Connection::releaseTableModel(TableModel* tableModel)
   tableModel->m_handle = invalidHandle;
 }
 
-void Connection::setTableModelRegion(TableModel* tableModel, int columnMin, int columnMax, int rowMin, int rowMax)
+void Connection::setTableModelRegion(TableModel* tableModel, uint32_t columnMin, uint32_t columnMax, uint32_t rowMin, uint32_t rowMax)
 {
   auto event = Message::newEvent(Message::Command::TableModelSetRegion);
   event->write(tableModel->handle());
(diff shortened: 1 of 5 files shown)


gfgit:
@reinder last 2 commits are not really related but help for long standing issues.
PropertyObjectEdit's m_property still does not have its `m_id` updated on change. Infact I use the value from the signal arg.
This is not a big problem as of commit 86c4685b69dae805b20b0df0092a0890ce7e550c objects can be retrieved without knowing the ID client side, so even with old ID you get things working.
To test, open a Block rail tile edit window with an input object set. Then edit input's ID and see it changing in Block window.

Another problem is editing Block Id which prefixes all block input sensor ids. This does not update them client side.
I've fixed this for `DecoderFunctions` but before applying it to all I would like your opinion on a more generic way of doing it.
For example passing Decoder directly to `DecoderFunctions::items` constructor and let it have a generic "connect to ID changed".
To be more correct, a SubObject changes ID if **any** of it's parents change ID, not only the top most. But since other parents are usually properties, they have fixed IDs like `DecoderFunctions` ID will always be "functions".

gfgit:
PropertyObjectEdit still does not update correctly every time. If object value changes then it follows object id changes but at first when window is opened it does not follow object id changes from the beginning

reinder:
For the parentId changed stuff we can do it similar to: `Object::loaded` and `Object::worldEvent`, SubObjects can propagate it to their SubObject's. Can you move the Id stuff to a seperate PR? The SubObject id's is indeed not yet working very well.

> Further testing seems to always work here now.

Did you try with Qt5 and Qt6 client?

gfgit:
> Did you try with Qt5 and Qt6 client?

No. I only build Qt6 on my laptop. I remember to have had this issue on Qt5 back then too but it would mean rebuilding all from start :S

gfgit:
Ok so here it remains only one commit which affects `TableWidget` so it will not fix #187 since it uses `InterfaceListWidget`.
But still this fixes other models using `TableWidget`

PR #169, opened 2024-09-17
XpressNet WIP

gfgit:
- Added `ATTRIBUTE_PACKED`
- Added chapters referring to Lenz docs

TODO:

I'm testing an USB interface (Oscilloscopio.it DCC-A500) which should be compatible with LI100f.
But I'm receiving a lot of garbage messages, some of them even pass the checksum test!
And I do not receive any message regarding locomotive status.

It seems that locomotive status feedback is sent only to the XpressNet device which is in control of that locomotive.
While on Z21 this kind of message is broadcast.
This is kind of weird, is Traintastic supposed to poll every Decoder for which a Throttle window is open, or which is part of a Train?
It would be a resource waste, there must be a better solution...

### DIFF
--- a/server/src/hardware/interface/xpressnetinterface.cpp
+++ b/server/src/hardware/interface/xpressnetinterface.cpp
@@ -290,27 +290,41 @@ bool XpressNetInterface::setOnline(bool& value, bool simulation)
           setState(InterfaceState::Error);
           online = false; // communication no longer possible
         });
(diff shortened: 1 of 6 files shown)


gfgit:
@reinder Even though I did not manage to make the my hardware work, this code is good so I think it can be merged!

reinder:
Thanks, I'll try to review asap.

PR #170, opened 2024-09-18
Train position tracking

gfgit:
Super ugly code just as proof of concept.

Now that #163 is quite working, I'm exploring new possibilities.
Knowing locomotives speed curves, (maybe combined together in a Train), we can now how much distance the Train has traveled.

Precision depends on how accurate is the speed curve file which gets imported for each locomotive.
I did some quick measures, without spending much time so accuracy could be improved a lot, and yet the result is already good.
In this small example:
- Take a locomotive with decoder set and speed curve loaded
- Put it alone in a new Train (so it uses raw curve data, since matching could introduce some rounding)
- Activate Train
- When you activate locomotive function F1 (then you can deactivate) Traintastic starts to track that Train until it travels for 20 centimeters (0.2 meters)
- You can make locomotive go as fast as you like, even stop it and restart (but not yet change direction)
- When Traintastic estimates traveled distance is >= 20 cm, locomotive F2 is enabled to tell you of the result.
- Tracking is deactivated until you (disable and) re-enable F1

From my tests the average error is of 2 centimeters which is quite acceptable for our purposes.

With this functionality we can for example:
- Release BlockPath when Train enters exit block and travels it's own length plus a safe margin, and of course once entrance block is freed.
- Make some automation based on position but without position sensors (like infrared or magnet) and at a finer precision that block sensors.
- Example play horn just before a level crossing or when entering/exiting a tunnel
- Play horn when crossing another Train on opposite track
- It's also useful to better control Train speed in future Automatic mode based on position, like start to brake only after 10 cm from yellow signal for whatever reason.

- Should tracker be an object which subscribes to a Train and then removes itself when reaching it's target?
- Multiple trackers could be active at same time for same Train
- Maybe we could use boost signals on properties but overhead should be kept to a minimum to keep time precision.

### DIFF
--- a/server/src/train/abstracttrainpositiontracker.cpp
+++ b/server/src/train/abstracttrainpositiontracker.cpp
@@ -3,6 +3,8 @@
 #include "../utils/almostzero.hpp"
 #include "../core/eventloop.hpp"
 
+#include <chrono>
+
 AbstractTrainPositionTracker::AbstractTrainPositionTracker(const std::shared_ptr<Train> &train)
   : m_train(train)
 {
@@ -48,7 +50,6 @@ void DeadlineTrainPositionTracker::trainSpeedChanged(double physicalSpeedMS)
     almostZero(currentTravelledMeters - targetTravelledMeters))
   {
     // We have reached target distance
-
     const double oldTarget = targetTravelledMeters;
 
     // Ask if we need to continue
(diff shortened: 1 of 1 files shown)


reinder:
Awesome 😎, need to check the code, but I really like it.

What will help I think to improve the accuracy is time stamping the input changes in the kernels, and send it with the change event, then we can eliminate the event loop delay. I can look into that.

I'll put this one on draft for now, we can then remove draft status once we are finalizing the last bits of a PR. (So I can easily see which ones I need to prioritize.)

Issue #185, opened 2025-02-19
Model railway simulator


reinder:
The model railway simulator can speed up software development, testing and makes demonstration/training is much easier.
Traintastic can then be used and tested without being connected to a real model railway. Due to the hardware interface simulation already present in Traintastic server it is even possible to see the command station communication.

📹  https://www.youtube.com/watch?v=5wof8e30d_U

reinder:
@gfgit if there is anything you want to work on, feel free to do it :) I'm currently working on getting adding simulator support on the traintastic-server side, most boiler plate code is ready, I just need to implement the command translation.

gfgit:
Hi Reinder,
what should we do for crosses and bridges?

Also doubke slip switches should go on their own or just be a pair of normal turnouts?

Can an occupancy sensor span across multiple track segments?
For example use same sensor for a group of turnouts which cannot hold multiple trains at same time.

reinder:
Hi @gfgit ,

Double/single slip, 3 way, Wye turnouts are missing just like bridges and crossings. Would be nice to add them. Feel free to create a PR.

Sensors are currently limited to segments, I already know how I want to fix that, but didn't have the time yet.

gfgit:
My idea was to create a sensor object with an id and then set sensor id inside each segment.

Another nice thing might be spline curves for flex tracks.
I think having only straight and curve would be too complex if one wants to closely replicate flex track.

reinder:
> My idea was to create a sensor object with an id and then set sensor id inside each segment.

Yes, something like that I have in mind :)

> Another nice thing might be spline curves for flex tracks.
> I think having only straight and curve would be too complex if one wants to closely replicate flex track.

That would be a nice addition, I just focused on the basics first to try if it would be something that can speedup the development of automatic train control :)

Yesterday I did some work on the XpressNet+simulator support, it now alse sends a "steal" message in case a the simulator throttle is adjusted. I have two command stations with XpressNet, so I can check if it behaves like the real hardware.

reinder:
@gfgit added shared sensor support. When you assign the same channel/address to multiple segments, they will share one sensor :)

gfgit:
I'll try to test it today!
Also I was thinking about adding point based train tracking.
Like adding a sensor to a track with position relative to track start position.
Then when train passes on that track it activates sensor for each vehicle.

Maybe it ca be a generic "object at position" so it can be resused in future to place signals and other stuff related to track.

[...]

PR #189, opened 2025-03-18
Simulator fixes

gfgit:
Hi, testing simulator with Z21 I've discovered powerOn state is not stored if change is triggered from simulator.
This means next `LAN_SYSTEMSTATE_DATACHANGED` will overwrite it.

Also I've noticed for now Emergency Stop message is not handled because simulator only has power on/off.
But I think it's fine for now. Also other interfaces like loconet doesn't seem to have separate Emergency stop command.


gfgit:
Last commit is about Clangd warning on missing direct includes, I've put it just in case...

reinder:
@gfgit thanks for the PR, nice improvement :)

The changes aren't compatible with Qt 6.4, those builds fail.

Could you change the `<qmath.h>` include to `<QMath>` (that's Qt style)

Emergency stop isn't implemented yet indeed, not all systems support that, when pressing `spacebar` all trains are stopped, but that currently sends speed=0 for each loco to the simulator client. (It is on the todo list, see #185)

gfgit:
Ups, I trusted IDE autocomplete :(
Will fix it!

PR #192, opened 2025-04-25
[simulator] fix start_point origin angle

gfgit:
This partially fixes turnouts starting from other segments origin.


gfgit:
Still if turnouts start from other turnouts "curve" side, and if curve angle are different, the other turnout is misplaced

PR #194, opened 2025-08-13
Bug simulator turnout curve

gfgit:
Fix for Simulator #185 

This fixes turnout with `"side": 2` and `"from_side": 2`
I'm sure there is a better way of doing this, like using the `curAngle` variable which is already dependent of `curve.angle` sign, but I didn't find it yet...


reinder:
Nice find, `curRotation` is wrong it should use `curAngle`, here a patch:
[fix-turnout-curve-start.patch.txt](https://github.com/user-attachments/files/21762484/fix-turnout-curve-start.patch.txt)

gfgit:
It works now. Thanks!

PR #195, opened 2025-08-17
Sim tooltips and double slip switches

gfgit:
Hi, this is an experiment branch.
I've added fast reload of JSON, so you can edit and see the new result instantly.
Then I've increased a bit the max zoom level and temporarily draw all segment in red because of better contrast on image background.
Image background is done with QPainter because I don't know OpenGL well enough...

Tooltips are super useful when building layout to adjust values. Code is still a bit messy.

Now about double slip switches:
- We parse the curve. How to calculate straight segments intersections?
- straightEnd does not support segments with 2 straights
- We now need to know from which side we came from to calculate to which side we are going to!

### DIFF
--- a/shared/src/traintastic/simulator/simulatorconnection.cpp
+++ b/shared/src/traintastic/simulator/simulatorconnection.cpp
@@ -30,6 +30,7 @@ SimulatorConnection::SimulatorConnection(std::shared_ptr<Simulator> simulator, b
   , m_connectionId(connId)
 {
     m_socket.set_option(boost::asio::ip::tcp::no_delay(true));
+    m_socket.set_option(boost::asio::socket_base::send_buffer_size(8192 * 2));
 }
 
 void SimulatorConnection::start()


reinder:
> Now about double slip switches:
> - We parse the curve. How to calculate straight segments intersections?

@gfgit I thought a bit about it during the day, see image:

image

`points[0]` is the origin, using the `angle` and `radius` we can calculate `points[1]`.

We also know the angle of both points, so we draw virtually two lines (red/orange) and calculate the intersection point `C`.
That kind of math is not really my cup of tea, but I promted ChatGPT, which came with this (untested, so it might need some adjustments/corrections):
```cpp
Point calcIntersection(Point pointA, float angleA, Point pointB, float angleB)
{
  // Direction vectors
  float dxA = std::cos(angleA);
  float dyA = std::sin(angleA);
  float dxB = std::cos(angleB);
  float dyB = std::sin(angleB);

  // Solve:
  // pointA + t * dirA = pointB + s * dirB
  // Rearranged: (pointB - pointA) = t * dirA - s * dirB

  const float det = dxA * (-dyB) - dyA * (-dxB);  // determinant of 2x2 system
  if(std::fabs(det) < 1e-6f)
  {
    return invalidPoint; // Lines are parallel or nearly parallel, no unique intersection.
  }

  const float rhsX = pointB.x - pointA.x;
  const float rhsY = pointB.y - pointA.y;

  // Solve for t using Cramer's rule
  const float t = (rhsX * (-dyB) - rhsY * (-dxB)) / det;

  Point intersection;
  intersection.x = pointA.x + t * dxA;
  intersection.y = pointA.y + t * dyA;
  return intersection;
}
```

Using the distance of `points[0]` to `C` we can calculate `points[2]`, same for `points[1]` and `points[3]`.

> - straightEnd does not support segments with 2 straights 

That needs to be modified, `TrackSegment` should be able to hold two straights -> `std::array<Straight, 2> straights;`
Similar to how it already works for curves.

We can also use those two straights for crossings, they can then be defined by a `length` and an `angle`.

> - We now need to know from which side we came from to calculate to which side we are going to!

Do you mean for the trains that are moving?

----

I'll take a look at the other improvements :)

reinder:
> Yes, we could add a while loop.
> About axle-counter behavior I think it's fine as is.
> Short tracks are just for adjusting layout shape so usually are part of same sensor as adjacent tracks.
> Thus whole sensor is already occupied when train passes over a short track.

You're right they are most likely share a sensor or don't have a sensor, so that will work for common case, i've added the loop, see c4f559288c7e95bc75322841ad6b13f1026a850b

> Position sensor with delay should restart the timer each time they get pressed.
> So if a train runs fast over it it could happen to get pressed with fisrt axle and then stay pressed until whole train has passed, then last axle will trigger 3s delay to get up again.
> If train is slow or has long wagons, the sensor might be able to get up after 3 seconds and the get pressed again by next axle.
> Delay can be configured as a JSON property of sensor.
> Zero delay will make it pulse on/off at each axle, so we cannot wait updateSensors() to trigger, we must switch it inside vehicle position update.
> I can also add delay on consuming application if you think it's better.
> Real word delay is mechanical and happens inside the sensor itself. It's basically an oil piston, going down it has a big pipe and to go back up it has a really small pipe so oil will flow slowly.

I prefer to the sensor trigger stuff in the `updateSensors()` function, a zero delay can then pulse for one simulator cycle. For the delay we can convert the time in seconds to cycles. Then every cycle we decrement it, once it reaches zero it becomes false again. This is easy to implement we don't need to have timer objects etc. it will however have a bit jitter due to the synchronization with the cycle time. But consuming software should be able to handle that, its more real world (model) train like :)

> For turnouts in real railways there are 2 kinds of them:
> - Ones which will change position if taken from wrong side
> - Ones which will make train deray and also break the turnout mechanism.
>
> Some turnouts in Italian railways can switch behavior based on reserved route.
> We could make it a property, auto switch on train passage if true, stop train and raise error if false.

We can add a global property to control that behavior and allow overriding it on turnout level. Do you need to be able to switch that behavior at runtime as well?

-----

Something else I thought about today, currently the simulator is part of the Traintastic installer. But as it becomes more and more an application that also can be used by other applications I think it is better to build a separate Windows installer for it.

gfgit:
Definitely we need a seprate installer for simulator!

Switching turnout behavior at runtime is not really a priority for me. It can be done leter if needed.

We need to move trains away from static data.
By polishing my PR we could add right click context menu, get the segment under mouse cursor and spawn/delete a train.
In future I could work on a timetable like feature.

gfgit:
Also I understand you want to keep it simple with updateSensor() and possible jitter is not a problem.
But this approach does not respect sensor trigger order.

For now is not much relevant as track circuits are longer than tick position delta so you trigger only one new sensor per vehicle move.
But it's weak and with new sensor types it can lead to inconsistencies.
For example near a Padua we had a small line segment, about 1 km or less, with a curious semi-automatic block circuit.
 think it was unique and not repeated elsewhere because circuit is a mess.
Anyway to free the block it wants to see track circuits get occupied in sequential order, then freed in sequential order and a position sensor to be triggered after forst 2 track circuits.
It doesn't matter how fast events happen but they must come in order otherwise the block remains in occupied stste even if train has already left.
Also axle couter sensors need to trigger for each axle and updateSensor() cannot know how many axle have passed on a sensor.
Converting delay into ticks is fine and adds a bit of randomness.

reinder:
> Definitely we need a seprate installer for simulator! 

Noted, I'll work on that.

> Switching turnout behavior at runtime is not really a priority for me.
> It can be done leter if needed.

Ok, then we can put that on the backlog :)

> We need to move trains away from static data. By polishing my PR we could add right click context menu, get the segment under mouse cursor and spawn/delete a train. In future I could work on a timetable like feature.

It depends, if we just use a static list of available trains (that are in the JSON), we just need to know in the `stateData` if the train is visible or not. Then when spawning a train you can only select one of the hidden trains (or do it via a command, would be useful for a "reset" like feature in your system I think.)

> Also I understand you want to keep it simple with updateSensor() and possible jitter is not a problem.
> But this approach does not respect sensor trigger order.

Within a tick is an issue indeed. For different trains it also is an issue (even if send directly), as trains are moved one by one, so the at the front of the list is always first. Even if another train would reach a sensor earlier due to speed/distance.

We can still increase the tick rate, as long as it can process everything with that time it should work. I'll add some logic to monitor that thread, so we can see much time it needs to calculate a tick vs the tick time.

For the drawing however we might need to skip frames, that's just visual and doesn't affect the simulation.

> Also axle couter sensors need to trigger for each axle and updateSensor() cannot know how many axle have passed on a sensor.
Or we send a `count` value for those, then it can go like: 1, 2, 4, 6, 7, 8 etc.

I think the simulator should be fully synced on the tick rate or fully be "live", then it also must fire triggers in the correct order is in respect to train speed/distance to the sensors. Mixed both techniques complicates it a lot and can become very messy and us much harder to understand and debug.

My preference is to keep the game/realtime system like tick rate and investigate if we can raise it.

reinder:
**Thicker track update**

You can use e.g. `glLineWidth(3.0f);` at the beginning of `drawTracks()`, at the end of the function set it back to `1.0f`.

If you zoom out far, the track becomes to thick, so we might need to scale it like the `trainWidth`, we can add a `trackWidth` property. Scale it for drawing with a minimum of `1.0f`.

reinder:
It might be an issue with the buffers, they are only 1024 bytes currently. Did you try to enlarge them? The write function just drops messages when there is no room in the write buffer. (That should be improved, there should be at least some kind of indication that it is happening.)

PR #198, opened 2025-10-05
Sim first batch

gfgit:
Hello,
these are a first set of changes rebased on current upstream.

Mainly UDP discovery and JSON comments.
I tried to increase TCP send butter to 8192 but problem is still present and async_write() doesn't seem to report error codes in callback so maybe the issue is elsewhere?

After this, if you agree, I removed completely the use of OpenGL to get thicker lines and to draw other objects like signals.


shared/src/traintastic/simulator/simulator.cpp
```diff
@@ -349,6 +353,8 @@ void Simulator::start()
     {
       if(m_serverEnabled)
       {
+        std::cout << "Starting server..." << std::endl;
+
```
reinder, line 357: please remove this, debug left over?

shared/src/traintastic/simulator/simulator.cpp
```diff
@@ -358,6 +364,23 @@ void Simulator::start()
...
+        if(ec)
+            assert(false);
+
+        m_socketUDP.bind(boost::asio::ip::udp::endpoint(boost::asio::ip::address_v4::any(), defaultPort), ec);
+        if(ec)
+        {
+            std::cout << "UDP cannot bind" << std::endl;
+        }
+
+        std::cout << ec.message() << " END" << std::endl << std::flush;
```
reinder, line 381: I suggest we change the return type from `void` to `std::error_code`, the start can return an error if something does wrong :) Then the callee can check it and report a proper error :)

shared/src/traintastic/simulator/simulator.cpp
```diff
@@ -367,10 +390,15 @@ void Simulator::start()
...
-  // ignore errors
+  std::cout << "Stopping server...";
+
+  try
+  {
+    m_socketUDP.close();
+  }
+  catch(...)
+  {
+  }
```
reinder, line 401: Why not use `ec` here?

shared/src/traintastic/simulator/simulator.cpp
```diff
@@ -358,6 +364,23 @@ void Simulator::start()
 
         m_acceptor.listen(5, ec);
 
+        m_socketUDP.open(boost::asio::ip::udp::v4(), ec);
```
reinder, line 367: I'd like to have discovery optional, we can add a `bool discoverable = false` parameter to the `start` function for that. In most cases discovery isn't needed I'd expect.

shared/src/traintastic/simulator/simulator.cpp
```diff
@@ -659,11 +698,60 @@ void Simulator::accept()
...
+        sendInitialState(*m_connections.rbegin());
         accept();
       }
     });
 }
 
+constexpr char RequestMessage[] = {'s', 'i', 'm', '?'};
+constexpr char ResponseMessage[] = {'s', 'i', 'm', '!'};
+
+void Simulator::doReceive()
```
reinder, line 710: Correct indent, must be 2 spaces :)

shared/src/traintastic/simulator/simulator.cpp
```diff
@@ -659,11 +698,60 @@ void Simulator::accept()
...
+                            // Swap bytes
+                            uint8_t b[2] = {};
+                            *reinterpret_cast<uint16_t *>(b) = response[2];
+                            std::swap(b[0], b[1]);
+                            response[2] = *reinterpret_cast<uint16_t *>(b);
+                        }
+
+                        std::cout << "UDP Sending to: " << m_remoteEndpoint.address().to_string()
+                                                        << " port: " << m_remoteEndpoint.port()
+                                                        << std::endl << std::flush;
```
reinder, line 738: remove `std::cout` stuff :)

shared/src/traintastic/simulator/simulator.hpp
```diff
@@ -243,7 +244,7 @@ class Simulator : public std::enable_shared_from_this<Simulator>
 
   StateData stateData() const;
 
-  void enableServer(bool localhostOnly = true, uint16_t port = 5741);
+  void enableServer(bool localhostOnly = true, uint16_t port = 5742);
```
reinder, line 247: Why is the default port changed?

gfgit: Hi, I've used 5471 for UDP discovery. But I can switch it to 5472 and put TCP back to original

shared/src/traintastic/simulator/simulator.cpp
gfgit: I need some help on this one...

shared/src/traintastic/simulator/simulator.cpp
gfgit: Done. Now we should add console argument to turn it on/off. For now it's still forced on because it needs testing.
Also another argument for localhost only server.

reinder: yes, command line options to control those would be great.

shared/src/traintastic/simulator/simulator.cpp
gfgit: Ok but the logic is run in the worker thread so returned error would be async.
What should I do?

reinder: worker thread should exit on error, the main thread then needs to check if the simulator is started or has an error. Those variables are accessed by two threads, so best is to use `std::atomic<>`:
```cpp
std::atomic<bool> m_started = false;
std::atomic<std::error_code> m_error;
```

When `start()` set `m_started` to `true` else set `m_error`.

The main thread then can check `(m_started || m_error)`, after the startup one of them must be true.

reinder:
@gfgit the handshake you added, is that a ping/keep alive?

gfgit:
Hi Reinder, what is the status on this?
I think I addressed all the remaining issues except returning std error_code on start.

In the meantime, buffer issue is solved by requesting specific channel at a time.

Also I've implemented square signals which go on right side of track and often have light arrow on top to clarify which track they refer to.
Instead the 2 small lights below are "advancemnt" or "call signal".
They are used when there are issues on apparatus so main signal remains red.

To increase automation I've created three new tracks objects:
1) Spawn train, when activated by a curcuit in my simulator, it will create a semi-random lengnht.

2) Remove train, directional.
Will erase train when it gets touched.

3) Change direction, when touched by an automatic train, this will invert travel direction. This node is less useful, I did it just for fun!
I plan to replace its use case with "normal stop locatio" tag on station platforms for commuter trains.

reinder:
CI/CD is running!

gfgit:
Hi Rinder, test are failing but compile fine.
It's failing on input tests. I don't think this is related to my PR. Maybe we should merge master to PR branch to have recent fixes?

reinder:
These input tests use the simulator, so something is different there.

For the test there are 4 sections with a sensor, a loco drives from 1 to 4 triggering events, the tests check if all event are fired in the correct order and no duplictes are received. It differs a bit per interface due to the protocols being different.

I'll try to have a look this weekend if I can spot what is causing the failures.

gfgit:
@reinder Tests were failing due to `Simulator::sendInitialState()` which was causing more input state changes than expected.
I've removed it in favor of `RequestChannel` message which can be sent by traintastic server and other simulator clients to get state of all sensor of a specified channel or all channels if invalid channel value is passed.

PR #203, opened 2025-11-17
User help box

kamil00110:
adds help info boxes for settigns. Can be used with:
Attributes::addHelp(centralUnitVersion, "terminJSON");


kamil00110 (commented):
just single clie additions. The main changes are in createwidget.cpp and objectwidget

reinder:
@kamil00110 thanks for the PR, I very much like this idea.

I'd like to suggest a few changes to improve the quality of the PR:
- Implement the *help* feature in the `InterfaceItemNameLabel` class, there are custom dialog that built the layout themself, by embedding it into the `InterfaceItemNameLabel` class it will work everywhere.
- The `InterfaceItem::helpText()` should work like:
  1. Check for *Help* attribute, check for *DisplayName* attribute if it is set add `/help` postfix, else use *classId* + *name* + `/help` as *term id*.
  2. Check if *term id* actualy exists in `Locale`, if it does return the translation, if not return an empty string.
- Squash all commits to a single commit adding the feature with a proper description.

PR #206, opened 2025-12-11
Add Attribute that allows to add help information to each displayname

kamil00110:
This implementation crates an small help icon behind each "DisplayName" Attribute. When hoverd over it shows an help message. It can be added to behind any DisplayName Attribute with `Attributes::addHelp(element, string);` the text can be set up in each language file. When the translation in some languages is missing then the help icon wont apear.
- Traintastic v0 3 0-help-box-cleanedup-385-a181b20b 11 12 2025 03_13_30

### DIFF
--- a/client/src/widget/createwidget.cpp
+++ b/client/src/widget/createwidget.cpp
@@ -136,52 +136,46 @@ QWidget* createWidget(AbstractProperty& baseProperty, QWidget* parent)
 
 QWidget* createWidget(Property& property, QWidget* parent)
 {
-    QWidget* widget = nullptr;
+  switch(property.type())
+  {
+    case ValueType::Boolean:
+      return new PropertyCheckBox(property, parent);
 
-    switch(property.type())
-    {
-        case ValueType::Boolean:
-            widget = new PropertyCheckBox(property, parent);
-            break;
+    case ValueType::Enum:
+      if(property.enumName() == "pair_output_action")
+      {
+        return new PropertyPairOutputAction(property, parent);
+      }
+      return new PropertyComboBox(property, parent);
 
-        case ValueType::Enum:
-            if(property.enumName() == "pair_output_action")
-                widget = new PropertyPairOutputAction(property, parent);
-            else
-                widget = new PropertyComboBox(property, parent);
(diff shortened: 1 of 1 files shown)


client/src/widget/interfaceitemnamelabel.hpp
kamil00110: Addition of the Help oftion to the label

client/src/widget/interfaceitemnamelabel.cpp
kamil00110: mainly the positioning and design of the help button

client/src/widget/createwidget.cpp
kamil00110: Handling of the creation of the help object

client/src/widget/object/objecteditwidget.cpp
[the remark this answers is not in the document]
reinder: Can you revert these changes in the PR?

server/src/core/attributes.hpp
```diff
@@ -97,7 +97,10 @@ struct Attributes
   {
     item.addAttribute(AttributeName::DisplayName, value);
   }
-
+  static inline void addHelp(InterfaceItem& item, std::string_view value)
+  {
+    item.addAttribute(AttributeName::Help, value);
+  }
```
reinder: please add empty line before and after the function

client/src/widget/createwidget.cpp
```diff
@@ -136,46 +138,78 @@ QWidget* createWidget(AbstractProperty& baseProperty, QWidget* parent)
...
+                        break;
+                    }
+                }
+                break;
+            }
+            current = current->parentWidget();
+        }
+    }
+
+    return widget;
```
reinder: This seems unnecessary, as the help stuff is now handled by the `InterfaceItemNameLabel`.

Besides that there is no guarantee that the widget is only used in a `QFormLayout`.

reinder (changes requested):
Nice work!, a few changes/improvements and I can merge it! :)

kamil00110:
I think that I fixed everything now.

PR #209, opened 2025-12-15
Added English help infos to some elements and interfaces

kamil00110:


shared/translations/en-us.json
kamil00110: Added probably the most inportant descriptions

PR #210, opened 2025-12-16
Fix Contributor names displayed as "Null" in Readme.md

kamil00110:


reinder:
Hi, this part of the README is updated by the [contributors-readme-action](https://github.com/akhilmhdh/contributors-readme-action). I just updated it to the latest version but it still display's *Null*, it should be fixed by https://github.com/akhilmhdh/contributors-readme-action/pull/77 but it isn't working.

PR #221, opened 2026-04-08
Xpressnet new messages

gfgit:
WIP


gfgit:
@reinder Still have to do more tests with ROCO multiMAUS and maybe other command stations if I can borrow them at the club.
The basic stuff is mostly done, do you like the structure?

gfgit:
I'm in doubt on listening for decoder property changes.
Decoder could change address and this is not handled currently.

reinder:
> I'm in doubt on listening for decoder property changes. Decoder could change address and this is not handled currently.

Think it is best to ignore the decoder address changed edge case, that shouldn't happen during normal operation.

I'm still thinking about merging `Decoder` and `RailVehicle` and change `DecoderController` to `RailVehicleController`, it is a big refactor, but will resolve these kind of issues, it will also make it possible to detect when a `Train` activates/deactivates an `RailVehicle`, that will make it possible to add/remove an loco address from the command station refresh list (or the LocoNet slot or the CBUS session etc.)

gfgit:
I had to listen to address changes as well.
If yoi create a new decoder while interface is online, the decoder briefly has address 0 and gets added to kernel poll list which trigger an assert on invalid address.
So now it skips polling 0 address decoders but updates their entry as soon as the address becomes valid so they can be properly polled.
We also have momentary function messages, which seems to set a function to be momentary or not.
JMRI uses these messages, manual doesn't say much about them but I'm pretty sure multiMAUS ignores this information so for now I think we can skip it.
With current polling structure it's easy to add them later ;)

One thing that is missing is turnout feedback but I've read on CBUS thread you are planning to rework this aspect so it's best to wait and do it in another PR.
Still non understood from manual but when sending a switch command I think shortly afterwards we should send it again with "active" set to false.
And it seems different turnouts messages cannot be interleaved, which is strange because I saw multiple turnouts switching simulataneously on some layouts...

reinder:
Hi,

Yes, I'm working in turnout feedback support. The basic are there already, need to do some further testing, I'd hope.tomhabe it ready within a week... :)

I remember I read in the Z21 specs that you may not interleave switch request, but in later versions they added a *Queued* flag. Can't remember if it was in the Xpress net specs, which document did you use as reference?

Greetings,
Reinder

gfgit:
Yes I read the Z21 manual too and saw the queue flag, which I added in older PR. But it seems queue has to do just with the ordering.
The active=true and then false must still be sent both.
Although it says newer switch decoder don't care about active flag, I think it's more about solenoid switches.

About Xpressnet I'm reading v4 manual from lenz website.
It's badly written and in german so I don't fully trust it. It has also some errors in message byte definitions so...

Point is from Z21 manual it seems  queue is done because of DCC protocol limitation of having only 1 switch active at a time.
If that's true (don't have a way to sniff DCC packets but would be cool!) then every command station should have this limitation.
Really don't know about it

gfgit:
Reading DCC specification there is no mention of switcj message ordering.

https://www.nmra.org/sites/default/files/s-9.2.1_2012_07.pdf

See Section D basic accessory

How does LocoNet handle active/non active flag?

reinder:
The issue here is in the command station, the DCC protocol is not the limitation.

Regarding LocoNet it also depends on the command station also I think, the Digikeijs DR5000 for example. If you send a switch on command, after about 200ms (if I remember correctly) the DR5000 will send a switch off command on LocoNet and probably on DCC as well.

The only thing we can do is test various command stations and based on that make a generic implementation and add settings if required to control the behavior. (Therefor LocoNet and XpressNet have a command station setting, so it easier for users to configure it correctly)

gfgit:
Rebased on current master to get #228 
- Added command station version query, use it to decide which messages to send
- Added command station info in interface window like done for Z21
- Added initial output state sync when switched from external throttle (here Roco addressing +4 doesn't seem to apply)

I've found this document about RocoNet which should be an XpressNet dialect [roconet-v1.6.1.pdf](https://github.com/Digital-MoBa/XpressNet/blob/main/roconet-v1.6.1.pdf)
Messages seem inconsistent, like "info" field set to 0x00 and "header" not reporting message size correctly.

Anyway how did you found Roco specific F13F28 message? I'm looking for a request message to poll for this reply, because multiMAUS XBus 3.0 does not support standard F13F28 messages.

I will check at the club if others have multiMAUS with newer firmware.

reinder:
Nice improvements!

> I've found this document about RocoNet which should be an XpressNet dialect [roconet-v1.6.1.pdf](https://github.com/Digital-MoBa/XpressNet/blob/main/roconet-v1.6.1.pdf)
Messages seem inconsistent, like "info" field set to 0x00 and "header" not reporting message size correctly.

This is the interface protocol, not XpressNet itself, the info 0x00 command seems to be the opcode for an XpressNet command. The interface also has RS-bus for feedback modules, which it can send to the PC.

> Anyway how did you found Roco specific F13F28 message?

Using an oscilloscope supporting UART decoding :) I measured the XpressNet signal on the 6 pin header and pressed some buttons on the MultiMAUS, then it was easy to figure out how it works. Implemented that in Traintastic and tested it to verify :)

gfgit:
@reinder Hi, I think the work here is ready for revision!

Summary of work done:
- Added tooltips to client's server log table, some messages are very long when printed and get truncated.
- Listen to Decoder address changes or function number changes in interface (let me know if code style is good)
- Send XpressNet messages directly from Lua script, useful for debugging but implemented badly. I can remove it or copy CBUS implementation.
- Implemented new messages, like central status, central version, turnout feedback
- Report Central version in Interface UI window
- Use central version for conditional message send, like turnout message old/new version, avoid functions >= F13 on old centrals.
- Implemented turnout feedback
- Implement locomotive status polling every 1000ms
  - Poll basic locomotive info (speed, direction, busy state, F0-F12)
  - Poll extended functions (F13-F28, needs XpressNet V3.6)
  - Poll extended functions (F29-F68, needs XpressNet V4.0)
  - Poll ROCO multiMAUS cumulative status (speed, direction, busy state, F0-F12, F13-F20)
- Better share code with Z21 messages and make common class LanXLocoBase

gfgit:
So far I've tested code with ROCO multiMAUS firmware V1.02, XpressNet V3.0
This means I can only test up to F20 but other messages are taken from documentation so it should be all correct.
I'm looking for someone with updated multiMAUS firmware or other XpressNet command station to further test it, I'd like to map command station reported ID to pretty names for UI.

I've discovered 2 new undocumented ROCO messages to query and get status of F13-F20 (Traintastic could be the first open source software to use them ;-)
In the reply message there are 2 unused bytes (maybe reserved for future use?) and a flag near the busy flag which is always on and cannot guess what it does...
I'm thinking to write a post in OpenDCC forum, I saw they have a table with all [XpressNet messages](https://www.opendcc.de/elektronik/opendcc/xpressnet_commands_e.html).

Another interesting aspect that confuses me a bit it the absence of polling between 2 multiMAUS (master and slave) both looking at same locomotive. In practice, the master sends messages when it drives a locomotive same way a slave would do, and the slave listens to every locomotive drive messages even if not addressed to itself, so it catches the state from other multiMAUS and does not need to poll to get locomotive state.
I don't know if this is also a ROCO custom extension or if it's possible also on Lenz command station.
The point is, PC interfaces cut all messages not addressed to them so computer will never get this extra drive commands and must poll for locomotive status (proof is JMRI does it too so must be the only way...)

gfgit:
I've implemented proper Lua scripting support, it was fun!
Also added the auto-checksum so it's easier to write scripts.

One thing I've noticed running a command station status request from script, is that we do not map well the power state, let me explain:
- World is in Power On + Run state, command station is in run state
- Press Power off, world goes in Power Off + Stop, command station goes to power off state
- Now press Power On, world goes into Power On + Stop, command station goes into Power Off + Emergency Stop state!!!

The state Power On + World stop is handled by XpressNet with emergency stop world message. The point is that this message only works if previously command station was in Run state.
To go from Power Off + Stop to Power On + Stop, it seems we need to resume operations and quickly emergency stop, because there is not a direct message for it.
So there is a tiny window when locomotives could run again but technically it's instantaneous.
Let me know if you want this approach, also Z21 might be affected by this bug.

gfgit:
Opened discussion on OpenDCC forum about undocumented ROCO messages: https://forum.opendcc.de/viewtopic.php?t=11171

reinder:
Hi!

Thanks for the PR, I'll review it. Need to find a block of time to do it, I'll try to do it this weekend.

Based on you comments above some quick notes:

> I've implemented proper Lua scripting support, it was fun!
Also added the auto-checksum so it's easier to write scripts.

Yes, that would be my request to implement it like in LocoNet and CBUS, I'd suggest to remove the checksum and autoChecksum completely for Lua, just like LocoNet and CBUS do, sending messages with incorrect checksum makes no sense.

Adding the `send` function also requires adding it in `manual/luadoc`, you can run `manual/build.py` using a proper Python venv to build it.

> *Power On + Stop* issue

Yeah this is a hard one, with every protocol/command station this is a bit different, it would be a lot easier if Traintastic just had two state: Power Off / Run. I choose three for a reason as many users control their turnouts via DCC, so Power On + Stop allow sending turnout command without driving trains.

What we could do when going from: PowerOff+Stop to PowerOn+Stop send an (E)Stop command to all known loco's before sending the resume command.

I really need to write something about this and the Traintastic HAL in general, that will make it much easier to verify if hardware behaves the same regardless of the used protocol (allthough there still can be hardware limitations, but we can document that in the manual.)

Huge thanks already for this PR!

server/src/hardware/interface/xpressnetinterface.cpp
```diff
@@ -201,6 +216,48 @@ void XpressNetInterface::decoderChanged(const Decoder& decoder, DecoderChangeFla
...
+  for(const auto& f : *decoder.functions)
+  {
+    if(f->number >= 29)
+      flags |= XpressNet::Kernel::Locomotive::Flags::HasF29F68;
+    else if(f->number >= 13)
+      flags |= XpressNet::Kernel::Locomotive::Flags::HasF13F28;
+
+    if((flags & bothFuncFlags) == bothFuncFlags)
+      break;
+  }
```
reinder: the logic for determining the *flags* is identical, best to move it to an function (can be local function in anonymous namespace at top of this cpp.)

server/src/hardware/interface/xpressnetinterface.cpp
```diff
@@ -241,6 +298,15 @@ bool XpressNetInterface::setOutputValue(OutputChannel channel, const OutputLocat
       m_kernel->setOutput(static_cast<uint16_t>(address), std::get<OutputPairValue>(value));
 }
 
+bool XpressNetInterface::send(std::vector<uint8_t> message, bool autoChecksum)
```
reinder: remove the `autoChecksum`, users should not be bothered with this (see LocoNet and CBus, that support this too)

server/src/hardware/interface/xpressnetinterface.cpp
```diff
@@ -465,3 +548,33 @@ void XpressNetInterface::updateVisible()
...
+    for(const auto& f : *decoder->functions)
+    {
+      if(f->number >= 29)
+        loco.flags |= XpressNet::Kernel::Locomotive::Flags::HasF29F68;
+      else if(f->number >= 13)
+        loco.flags |= XpressNet::Kernel::Locomotive::Flags::HasF13F28;
+
+      if((loco.flags & bothFuncFlags) == bothFuncFlags)
+        break;
+    }
```
reinder: this can also be done using a generic function, as mentioned above

server/src/hardware/protocol/xpressnet/iohandler/simulationiohandler.cpp
```diff
@@ -47,17 +47,123 @@ bool SimulationIOHandler::send(const Message& message)
...
+              // Simulate external change when setting function 68
+              m_kernel.ioContext().post(
+                [this, address=funcInstruction.address()]()
+                {
+                  Locomotive& loco2 = m_locomotives[address];
+                  loco2.info.setBusy(true);
+                  loco2.info.updateChecksum();
+                  m_kernel.receive(LocomotiveBusy(address));
+                });
+            }
```
reinder: Is this a testing left over or is it there for a reason?

server/src/hardware/protocol/xpressnet/kernel.cpp
```diff
@@ -32,13 +34,147 @@
 
 namespace XpressNet {
 
+void Kernel::postQuery(const Utils::PendingQuery &query)
+{
+  assert(query.address >= shortAddressMin && query.address <= longAddressMax);
```
reinder: In most Kernels I've added `assert(isKernelThread());`,  `assert(isEventLoopThread());` on top of every function, this is really helpful for catching threading issues. Can you add these to the new functions (and existing if you like)?

server/src/hardware/protocol/xpressnet/kernel.hpp
```diff
@@ -42,19 +43,37 @@ class OutputController;
...
+enum HardwareType : uint8_t;
 
 class Kernel : public ::KernelBase
 {
   public:
     static constexpr uint16_t inputAddressMin = 1;
     static constexpr uint16_t inputAddressMax = 2048;
     static constexpr uint16_t accessoryOutputAddressMin = 1;
-    static constexpr uint16_t accessoryOutputAddressMax = 1024;
+    static constexpr uint16_t accessoryOutputAddressMax = 2048; // TODO: depends on XBus version (1024 up to 3.6)
```
reinder, line 54: using 2048 for this is fine, as that is the theoretical limit.

Nice to have would be to warn the users after the XBus version is read and is not 4.0 while there is a output configured using an address above 1024. (the XpressNetInterface should do that)

server/src/hardware/protocol/xpressnet/kernel.hpp
```diff
@@ -42,19 +43,37 @@ class OutputController;
...
+    struct Locomotive
+    {
+      uint16_t address = 0;
+
+      enum Flags : uint8_t
+      {
+        None = 0x0,
+        OwnedByXBus = 0x1,
+        HasF13F28 = 0x2,
+        HasF29F68 = 0x4
```
reinder: These are used as bits in a mask, I prever to write them as: `1 << 0`, `1 << 1`, `1 << 2` etc. so it is clear it is used as mask

server/src/hardware/protocol/xpressnet/kernel.hpp
```diff
@@ -97,6 +116,32 @@ class Kernel : public ::KernelBase
...
+    {
+        XNet_2_3 = 0x23,
+        XNet_3_0 = 0x30,
+        XNet_3_6 = 0x36,
+        XNet_3_8 = 0x38,
+        XNet_4_0 = 0x40
+    };
+
+    CentralVersion m_centralVersion = CentralVersion::XNet_3_0;
+    CentralVersion m_centralVersionEventLoop = CentralVersion::XNet_3_0;
```
reinder: isn't this actually the XBus/XpressNet protocol version?

server/src/hardware/protocol/xpressnet/messages.hpp
```diff
@@ -120,11 +172,64 @@ struct EmergencyStop : Message
...
   EmergencyStop()
   {
-    header = 0x81;
+    header = BC_STOPPED;
   }
 } ATTRIBUTE_PACKED;
 static_assert(sizeof(EmergencyStop) == 3);
 
+// 2.7.1 Software version reply (OLD! up to Central version 2.3)
+struct CentralVersionReplyOLD : Message
```
reinder: name it `CentralVersionReplyV2` ? or is it also V1?

server/src/hardware/protocol/xpressnet/messages.hpp
```diff
@@ -540,13 +985,337 @@ struct setFunctionStateGroup : LocomotiveInstruction
...
+
+  inline void setFunctionBase(uint8_t index, bool value)
+  {
+    assert(index >= 0 && index <= 12);
+    if(index == 0)
+    {
+      if(value)
+        functions1 |= flagF0;
+      else
+        functions1 &= ~flagF0;
```
reinder, line 1103: Tip: the is a utility function for this: `setBit`, see [`server/src/utils/bit.hpp`](https://github.com/traintastic/traintastic/blob/master/server/src/utils/bit.hpp)

server/src/hardware/protocol/xpressnet/utils.hpp
```diff
@@ -122,6 +189,31 @@ constexpr uint8_t fromBCD(uint8_t value)
...
+
+static constexpr uint8_t xbusVersionMajor(uint8_t versionHex)
+{
+  return (versionHex >> 4) & 0x0F;
 }
 
+static constexpr uint8_t xbusVersionMinor(uint8_t versionHex)
+{
+  return versionHex & 0x0F;
+}
```
reinder, line 225: Just an idea, define a `struct`:
```cpp
struct XBusVersion
{
  XBusVersion() = default;

 constexpr XBusVersion(uint8_t versionBCD)
    : m_versionBCD{versionBCD}
  {
  }

  constexpr  XBusVersion(uint8_t versionMajor, uint8_t versionMinor)
    : m_versionBCD(((versionMajor & 0x0F) << 4) | (versionMinor & 0x0F))
  {
  }

  uint8_t major() const
  {
    return (m_versionBCD >> 4) & 0x0F;
  }

  uint8_t minor() const
  {
    return m_versionBCD & 0x0F;
  }

  auto operator<=>(const XBusVersion&) const = default;

private:
  uint8_t m_versionBCD;
};
```

Then you can also do comparisons like `if(version >= XBusVersion(3, 6))` :)

reinder (changes requested):
Really nice work! Just a few notes :)

server/src/hardware/interface/xpressnetinterface.cpp
gfgit: Done

server/src/hardware/interface/xpressnetinterface.cpp
gfgit: Removed autoChecksum. Now it's always autochecksummed, really helpful for debugging complex messages

server/src/hardware/interface/xpressnetinterface.cpp
gfgit: Done as above

server/src/hardware/protocol/xpressnet/iohandler/simulationiohandler.cpp
gfgit: Yep, now removed!

server/src/hardware/protocol/xpressnet/kernel.cpp
gfgit: Done everywhere

server/src/hardware/protocol/xpressnet/kernel.hpp
gfgit: Done

server/src/hardware/protocol/xpressnet/kernel.hpp
gfgit: Yes, renamed

server/src/hardware/protocol/xpressnet/messages.hpp
gfgit: Done

server/src/hardware/protocol/xpressnet/messages.hpp
gfgit: I honestly don't like setBit() because of template parameter, this pattern is used a lot elsewere.
`flagF0` would turn in a `setBit<5>(functions1)` ?

server/src/hardware/protocol/xpressnet/utils.hpp
gfgit: While I like the idea, I think it's overengineered XD (I usually love overengineer stuff)
We already have fromBCD/toBCD helpers, used in Z21 kernel.
I would like an unified solution for both Z21 and XpressNet

server/src/hardware/protocol/xpressnet/kernel.hpp
gfgit: Done. It's 3.8 the version introducing 2048 switches

server/src/hardware/protocol/xpressnet/messages.hpp
reinder: It is extensively used in newer hardware implementation, e.g. here: https://github.com/traintastic/traintastic/blob/master/server%2Fsrc%2Fhardware%2Fprotocol%2Fcbus%2Fmessages%2Fcbusenginemessages.hpp#L274

reinder:
@gfgit I've added the documentation for the new XpressNet `send()` Lua function, can you have a look at the `manual/docs/en/appendix/xpressnet.md` manual page, there are some more things supported now due to our work :)

server/src/hardware/protocol/xpressnet/utils.hpp
reinder: We could name it e.g. `BCDVersion` and put it in `utils` so it is usable by both, but it is just a nice to have :)

gfgit:
Cool, I'll keep it for reference if I need to add other things to manual in the future. Thanks!

PR #223, opened 2026-04-21
Add a script to build an x86_64 AppImage

DL7BJ:
Create folder package/appimage and a script, which build a Traintastic AppImage for Linuix x86_64.

The resulting AppImage bundles both the traintastic server and client.

On startup, the server is launched first, followed shortly by the client.
When the client exits, the server is automatically stopped as well.

### DIFF
--- a/client/src/dialog/screenshotdialog.cpp
+++ b/client/src/dialog/screenshotdialog.cpp
@@ -2,7 +2,7 @@
  * This file is part of Traintastic,
  * see <https://github.com/traintastic/traintastic>.
  *
- * Copyright (C) 2025 Reinder Feenstra
+ * Copyright (C) 2025-2026 Reinder Feenstra
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
@@ -51,7 +51,7 @@
 #include "../widget/objectlist/objectlistwidget.hpp"
 #include "../widget/objectlist/interfacelistwidget.hpp"
 #include "../widget/objectlist/throttleobjectlistwidget.hpp"
-#include "../widget/outputmapwidget.hpp"
+#include "../widget/iomapwidget.hpp"
 #include "../widget/tile/tilewidget.hpp"
 #include "../widget/throttle/throttlebutton.hpp"
 #include "../widget/throttle/throttlewidget.hpp"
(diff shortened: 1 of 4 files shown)


reinder:
Moin Tom,

Nice going to give this a try.

Can we build images for armhf and arm64 as well?
It would be nice to have the version number in the filename as well.

I like the idea of a combined image, that will be useful for users, can we build seperate images as well?

Greetings,
Reinder

p.s. for PR's it is best to create a branch on you fork, now it is much harder to sync, it will require a merge or rebase.

DL7BJ:
Moin Reinder, 

> Can we build images for armhf and arm64 as well? 

Yes, sure. But you need a arm device or qemu/docker to build the appimage. The simple way is to use a Raspberry for it. If you want both, 32-bit and 64-bit than you need a 32-bit and a 64-bit system to build the appimage with the right libs. I habe some Pi 4, I will build an appimage and a full Pi Image with Traintastic and MMC. 

> It would be nice to have the version number in the filename as well.

Is there one file, where the script could read the version number? It would also be nice to copy the executables in one specific folder, to upload the new files to your webserver. 

> I like the idea of a combined image, that will be useful for users, can we build seperate images as well?

Yes, of course. The combined image is the simplest way to run Traintastic without installation. 

> p.s. for PR's it is best to create a branch on you fork, now it is much harder to sync, it will require a merge or rebase.

Next time 😏 I have had much trouble with the fork, because the link to both submodules where absolute instead relative, so I got a 404 from github. And then it was to late 😏

Greetings, Tom

realm-hue:
Maybe consider docker images ?
Gives a bit more abstraction?

DL7BJ:
> Maybe consider docker images ? Gives a bit more abstraction?

Docker is certainly useful—especially for administrators ;-)

However, the goal is to make Traintastic usable for model railroad enthusiasts with limited knowledge of computers and operating systems. In practice, launching an AppImage is much easier for these users than managing Docker.

For the Raspberry Pi, a fully preconfigured system image is planned. The idea is simple: write the image to an SD card, power up the Raspberry Pi, and start using the digital model railway control right away.

reinder:
I prefer AppImage over Docker too. Docker is to much for the average user.

@DL7BJ I've two RPi4's one 32bit and one 64bit, they build the Raspbian binaries and deb packages of Traintastic, so they can build the AppImage as well.

The version number can be read from `shared/traintastic.cmake` and some env. vars, this logic is that file too. That needs to be added to the script or maybe rewrite it in Python :)

DL7BJ:
> The version number can be read from `shared/traintastic.cmake` and some env. vars, this logic is that file too. That needs to be added to the script or maybe rewrite it in Python :)

I hate Python! bash, awk, sed is enough for such simple things and if I need more, than Perl. Yes, I am old ;-)

realm-hue:
> > The version number can be read from `shared/traintastic.cmake` and some env. vars, this logic is that file too. That needs to be added to the script or maybe rewrite it in Python :)
> 
> I hate Python! bash, awk, sed is enough for such simple things and if I need more, than Perl. Yes, I am old ;-)

Well I hate learning new languages and lua ends on my “hate” list. Was convinced to pivot from Perl to python.
Plus the beginner level c/c++ 
Never enjoyed Java except for the black/drinkable variety 
;-)

but I may have to learn lua after all to make traintadtic sing and dance 🤣

PR #232, opened 2026-05-14
Documentation

DL7BJ:
Complete the translation to German.


reinder:
I've added support for building the German manual as well to `build.py`, moved language specific string to a new json file: `/manual/terms/de.json`. Created the translation with AI, can you check it for correctness?

DL7BJ:
> I've added support for building the German manual as well to `build.py`, moved language specific string to a new json file: `/manual/terms/de.json`. Created the translation with AI, can you check it for correctness?

That looks good! I tested the build script too, no errors.

PR #233, opened 2026-05-29
Waiting for socket established in server.cpp for slow computer

DL7BJ:
Moin Reinder,

I add a wait loop for a stable socket. 

Greetings, Tom

### DIFF
--- a/server/src/network/server.cpp
+++ b/server/src/network/server.cpp
@@ -279,6 +279,9 @@ Server::Server(bool localhostOnly, uint16_t port, bool discoverable)
   , m_manualPath{getManualPath()}
 {
   assert(isEventLoopThread());
+  bool socketStable = false;
+  int retries = 0;
+  const int maxRetries = 100;
 
   boost::system::error_code ec;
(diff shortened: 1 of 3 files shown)


reinder:
Moin Tom,

Is this a fix for the [server not connectable](https://discourse.traintastic.org/t/installation-von-traintastic-0-4-0-unter-windows-10-funktioniert-nicht/133) issue?

Did you find somewhere this could be the issue?

Greetings,
Reinder

reinder:
I've started the pipeline, so we get a windows build we can use for testing, if this is the fix I'll merge it after a bit of syntax polishing. :)

DL7BJ:
Moin Reinder,

> Is this a fix for the [server not connectable](https://discourse.traintastic.org/t/installation-von-traintastic-0-4-0-unter-windows-10-funktioniert-nicht/133) issue?
> 
> Did you find somewhere this could be the issue?

Yeah! I just posted my suspicions in the forum. It's an attempt, I suspect it's just a timing issue.

Greetings, Tom

reinder:
Looking at this again, it won't work, before the *stable* check, `local_endpoint()` is used for the *listening at* log message, so if it isn't valid, it will throw an exception, see [here](https://www.boost.org/doc/libs/latest/doc/html/boost_asio/reference/basic_socket_acceptor/local_endpoint/overload1.html)

Closing the PR.

